PR #1, opened 2020-06-29
Disable debug mode on CI

marek1840:


driver/sources/src/main/scala/com/virtuslab/ideprobe/ide/intellij/DebugMode.scala
```diff
@@ -1,12 +1,19 @@
...
 object DebugMode {
-  private val suspend: Boolean = false
-  private val port: Int = 5005
+  private val enabled: Boolean = env("IDEPROBE_DEBUG", "false").toBoolean
+  private val suspend: Boolean = env("IDEPROBE_DEBUG_SUSPEND", "false").toBoolean
+  private val port: Int = env("IDEPROBE_DEBUG_PORT", "5005").toInt
 
-  def vmOption: String = {
+  def vmOption: Seq[String] = {
     val suspendOpt = if (suspend) "suspend=y" else "suspend=n"
```
lukaszwawrzyk, line 9: could be inside if

PR #3, opened 2020-07-02
Cross-build against scala versions 2.12 and 2.13

marek1840:


api/src/main/scala/org/virtuslab/ideprobe/log/IdeaLogParser.scala
```diff
@@ -38,10 +39,14 @@ object IdeaLogParser {
...
+    !matches(errorLineRegex, line) && matches(anyLogLineRegex, line)
   }
 
   private def isNotErrorLine(string: String): Boolean = {
-    !errorLineRegex.matches(string)
+    !matches(errorLineRegex, string)
+  }
+
+  private def matches(regex: Regex, string: String): Boolean = {
+    regex.pattern.matcher(string).matches()
```
lukaszwawrzyk, line 50: I am wondering what happened here. So it seemed it was enough to just call regex.matches(string) and now we have matches(regex, string) with different implementation. Why so?

marek1840: sadly, `Regex::matches` is not available in scala 2.12

PR #6, opened 2020-07-06
Handle optional overriding plugins by appending to plugin list

lukaszwawrzyk:


driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJFactory.scala
```diff
@@ -28,28 +31,44 @@ final class IntelliJFactory(dependencies: DependencyProvider, val config: Driver
   private def installIntelliJ(version: IntelliJVersion, root: Path): Unit = {
     println(s"Installing $version")
     val file = dependencies.fetch(version)
-    unpackTo(file, root)
+    toArchive(file).extractTo(root)
     root.resolve("bin/linux/fsnotifier64").makeExecutable()
     println(s"Installed $version")
   }
 
   private def installPlugins(plugins: Seq[Plugin], root: Path): Unit = {
```
tpasternak, line 39: I'd comment that the method distinguishes plugins only by comparison of root files in the archive

PR #7, opened 2020-07-08
Fix findClass out of read action

lukaszwawrzyk:


olafurpg:
I'm curious, does it make sense to cross-publish if the Scala plugin is implemented in 2.13?

PR #8, opened 2020-07-10
Support vcs roots

lukaszwawrzyk:


driver/bindings/junit/src/main/scala/org/virtuslab/ideprobe/DSL.scala
```diff
@@ -13,19 +13,19 @@ final class RunningIntelliJFixture(
...
+  def runShell(args: Seq[String]): CommandResult = Shell.run(path, args: _*)
 
   def config: Config = fixture.config
 
   def intelliJPaths: IntelliJPaths = installedIntelliJ.paths
 
-  def run[A](action: RunningIntelliJFixture => A): A = {
-    val running = fixture.startIntelliJ(workspace, installedIntelliJ)
-    val data = new RunningIntelliJFixture(workspace, running.probe, config, intelliJPaths)
+  def runIntellij[A](action: RunningIntelliJFixture => A): A = {
```
marek1840, line 26: maybe `runIDE`? it's shorter and more generic

driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
```diff
@@ -3,5 +3,5 @@ package org.virtuslab.ideprobe.dependencies
 final case class IntelliJVersion(build: String) extends AnyVal
 
 object IntelliJVersion {
-  val Latest = IntelliJVersion("202.5792.28-EAP-SNAPSHOT")
+  val Latest = IntelliJVersion("202.6250.13-EAP-SNAPSHOT")
```
marek1840, line 6: what about the version of the plugin itself in `build.sbt`? Should it stay unchanged?

driver/bindings/junit/src/main/scala/org/virtuslab/ideprobe/DSL.scala
lukaszwawrzyk: I'd keep intellij for now. We have `Intellij` all over the code/api.

driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
lukaszwawrzyk: I used buildinfo to have it in one place

PR #9, opened 2020-07-10
Describe the driver purpose and configuration

marek1840:
This describes how to configure the ide probe. We still need to describe how to actually use the probe to write tests.

### DIFF
--- a/docs/driver.md
+++ b/docs/driver.md
@@ -1,0 +1,31 @@
+### Driver
+
+Driver controls launching the IDE  
+
+#### launch Command
+`driver.launch.command = ["idea"]`
+
+The command used to launch the IDE. By default, the driver uses the `idea` command, but user can specify a custom command. Note, that the `idea` is available on the `PATH`, if the user wants to wrap it in some custom logic. 
+##### launch Timeout
+`driver.launch.timeout = "30 seconds"`
+
+By default, the driver waits `30 seconds` for the IDE to connect the probe. After exceeding this time, the whole test fails.  
+#### Automatic checks
+```
+driver.checks {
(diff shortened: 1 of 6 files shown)


docs/driver.md
```diff
@@ -0,0 +1,31 @@
...
+The command used to launch the IDE. By default, the driver uses the `idea` command, but user can specify a custom command. Note, that the `idea` is available on the `PATH`, if the user wants to wrap it in some custom logic. 
+##### launch Timeout
+`driver.launch.timeout = "30 seconds"`
+
+By default, the driver waits `30 seconds` for the IDE to connect the probe. After exceeding this time, the whole test fails.  
+#### Automatic checks
+```
+driver.checks {
+    errors = false
+    freezes = false
```
lukaszwawrzyk, line 17: I think currently we have freezes = true, but perhaps it should be as in docs

docs/workspace.md
```diff
@@ -0,0 +1,37 @@
+#### Workspace
+
+User can use workspace to predefine any data required by a single test case.
+     
+It can be defined as one of:
+
+1. Directory on the classpath
+
+`probe.workspace.path = "classpath://foo/bar"`
```
lukaszwawrzyk, line 9: we do single `/` here, not sure if it works with double

docs/workspace.md
```diff
@@ -0,0 +1,37 @@
...
+2. Directory within a jar
+ 
+`probe.workspace.path = "jar://foo.jar!/bar/baz"`
+
+3. Github repository
```
lukaszwawrzyk, line 14: I think we also support file://

PR #10, opened 2020-07-16
Couple of small improvements

lukaszwawrzyk:
- Handle multiple "build steps" (one triggered build can result in multiple separate compilation runs).
- Change FileRef to use Path rather than String
- Add `includeNonProjectItems` parameter to NavigationQuery
- Add 2 fixtures to run shared intellij/workspace for whole test suite
- Extend probe configuration lookup logic to support companion objects and TestName.conf (aditionally to TestName/ideprobe.conf)
- Disable failing on freezes by default (needs to be explicitly enabled)
- `CommandResult.out` throws if command failed (similarly to .!! in scala.process._) We often just run a shell command and take the output, in case command fails, we get unexpected or empty output and the problem is hidden. Output can be always read with `outSafe`, yet we don't even need to e.g. log it, as it is always forwarded to console.

### DIFF
--- a/driver/bindings/junit/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
+++ b/driver/bindings/junit/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
@@ -1,0 +1,75 @@
+package org.virtuslab.ideprobe
+
+import java.nio.file.Path
+
+import org.junit.AfterClass
+import org.junit.BeforeClass
+import org.virtuslab.ideprobe.ide.intellij.InstalledIntelliJ
+import org.virtuslab.ideprobe.ide.intellij.RunningIde
+
+trait RunningIntellijPerSuite {
+
+  private var workspace: Path = _
+  private var installed: InstalledIntelliJ = _
+  private var running: RunningIde = _
+  private var runningIntelliJFixture: RunningIntelliJFixture = _
+
+  final def intelliJ: RunningIntelliJFixture = runningIntelliJFixture
+
+  protected def baseFixture: IntelliJFixture
+
+  protected def beforeAll(): Unit = ()
+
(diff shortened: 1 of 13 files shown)


driver/bindings/junit/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.BeforeClass
+import org.virtuslab.ideprobe.ide.intellij.InstalledIntelliJ
+import org.virtuslab.ideprobe.ide.intellij.RunningIde
+
+trait RunningIntellijPerSuite {
+
+  private var workspace: Path = _
+  private var installed: InstalledIntelliJ = _
+  private var running: RunningIde = _
+  private var runningIntelliJFixture: RunningIntelliJFixture = _
```
marek1840, line 15: maybe lazy vals ?

probePlugin/src/main/scala/org/virtuslab/handlers/Builds.scala
```diff
@@ -76,11 +75,17 @@ object Builds extends IntelliJApi {
     connection
   }
 
+  // We might get more than one build result form one build. buildResult is called after
```
marek1840, line 78:
```suggestion
  // We might get more than one build result from one build. buildResult is called after
```

driver/bindings/junit/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
lukaszwawrzyk: These are meant to be static later, kind of annoying but @BeforeClass and @AfterClass must be static, so the trait will be mixed in into an object. I don't think it is safe to allow single initialization in such case.

PR #14, opened 2020-07-22
Expose module sources and dependencies

marek1840:


api/src/main/scala/org/virtuslab/ideprobe/protocol/Endpoints.scala
```diff
@@ -29,6 +29,11 @@ object Endpoints {
   val ListOpenProjects = Request[Unit, Seq[ProjectRef]]("projects/all")
   val Messages = Request[Unit, Seq[IdeMessage]]("messages")
   val ModuleSdk = Request[ModuleRef, Option[String]]("module/sdk")
+  val ModuleDependencies = Request[ModuleRef, Seq[ModuleRef]]("module/dependencies")
+  val ModuleSources = Request[ModuleRef, Seq[Path]]("module/sources")
```
lukaszwawrzyk, line 33: Maybe we could just have one endpoint for content roots of module? Also content roots are already in basic project model, maybe it could be extended with types

marek1840: Now, the contents root in the project model are a map of ContentRoot -> Set of paths

probePlugin/src/main/scala/org/virtuslab/ProbePluginExtensions.scala
```diff
@@ -0,0 +1,34 @@
...
+
+    def contentRoots(kind: ContentRoot): (ContentRoot, Set[Path]) = {
+      val rootType = kind match {
+        case ContentRoot.MainSources   => SOURCE
+        case ContentRoot.MainResources => RESOURCE
+        case ContentRoot.TestSources   => TEST_SOURCE
+        case ContentRoot.TestResources => TEST_RESOURCE
+      }
+
+      kind -> roots.getSourceRoots(rootType).asScala.map(_.toNioPath).toSet
```
lukaszwawrzyk, line 29: Let's use VFS.toPath here. toNioPath is a new method and if you run probe with older intellij it fails.

probePlugin/src/main/scala/org/virtuslab/ProbePluginExtensions.scala
```diff
@@ -0,0 +1,34 @@
...
+object ProbePluginExtensions {
+  final implicit class ModuleExtension(module: Module) {
+    def toRef: ModuleRef = {
+      val project = module.getProject
+      ModuleRef(module.getName, ProjectRef(project.getName))
+    }
+
+    def dependencies: Array[Module] = roots.getDependencies
+
+    def contentRoots(kind: ContentRoot): (ContentRoot, Set[Path]) = {
```
lukaszwawrzyk, line 21: this is quite odd api.
```scala
def returnX(a: A): (A, X) = {
  a -> getX
}
```
It feels like creating tuple with param belongs outside of it

api/src/main/scala/org/virtuslab/ideprobe/protocol/Project.scala
```diff
@@ -1,9 +1,52 @@
...
+                  kind: Option[String])
+
+object Module extends ConfigFormat {
+  implicit val reader: ConfigReader[Module] = exportReader[Module].instance
+  implicit val writer: ConfigWriter[Module] = exportWriter[Module].instance
+
+  import pureconfig._
+  import pureconfig.configurable._
+
+  implicit val contentRootsReader: ConfigReader[Map[ContentRoot, Set[Path]]] =
```
lukaszwawrzyk, line 30: wouldn't it be generated automatically (in some a bit less readable form, but still?

marek1840: Unfortunately it fails because it cannot convert `ContentRoot` to string and vice versa (even with manually specifying readers and writers for `ContentRoot`). It probably does not know it is being converted to string and not an arbitrary object.

lukaszwawrzyk: Oh, that's likely a limitation as in hocon map keys must be strings, and general mechanism of derivation would create objects for these (like `{ type = "MainSources" }`, so it couldn't be used as a key...

probePlugin/src/main/scala/org/virtuslab/ProbePluginExtensions.scala
marek1840: It no longer creates a tuple here, but rather on the call site

PR #17, opened 2020-07-30
Add basic project import test examples

MichalBednarz:


driver/examples/src/test/scala/org/virtuslab/ideprobe/Example.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  private val fixture = IntelliJFixture(
+    version = IntelliJVersion("202.5792.28-EAP-SNAPSHOT"),
+    plugins = List(Plugin("org.intellij.scala", "2020.2.584", Some("nightly")))
+  )
+
+  private val config = Config.fromClasspath("SbtProject/ideprobe.conf")
+  private val fixtureFromConfig = IntelliJFixture.fromConfig(config)
+
+  def testModuleTestSuiteRun(moduleName: String, intelliJ: RunningIntelliJFixture): Unit = {
```
marek1840, line 26: name could be improved, like `runsAllModuleTests`

driver/examples/src/test/scala/org/virtuslab/ideprobe/ModulePresent.scala
```diff
@@ -0,0 +1,23 @@
+package org.virtuslab.ideprobe
+
+import org.junit.{Assert, Test}
+
+class ModulePresent extends ModuleTest {
```
marek1840, line 5: class name is not very descriptive, Perhaps, we could move the `run` method into the `ModuleTest` class and call it `verifyModulePresent` while at the same time inlining the `testPresentModules` method from this class? As a result we would have one less class with a more descriptive test name

driver/examples/src/test/scala/org/virtuslab/ideprobe/ModuleTestRun.scala
```diff
@@ -0,0 +1,24 @@
+package org.virtuslab.ideprobe
+
+import org.junit.{Assert, Test}
+import org.virtuslab.ideprobe.protocol.{JUnitRunConfiguration, ModuleRef}
+
+class ModuleTestRun extends ModuleTest {
```
marek1840, line 6: this also can become a single method in the `ModuleTest` class called `runsAllTestsInModule`

driver/bindings/junit/src/main/scala/org/virtuslab/ideprobe/DSL.scala
```diff
@@ -9,6 +9,7 @@ final class RunningIntelliJFixture(
     val workspace: Path,
     val probe: ProbeDriver,
     val config: Config,
+    val test: TestConfig,
```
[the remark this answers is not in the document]
MichalBednarz: should I implement the implicit `ConfigReader` for the `TestConfig` or alternatively simply utilising something like the `get[Seq[String]]("test.modules")` on the final `Config` will suffice?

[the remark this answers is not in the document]
marek1840: Implicit reader should be OK. It will simplify the call-site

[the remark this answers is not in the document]
lukaszwawrzyk: Each test usually has different kind properties, and config is designed and used in more ad-hoc way.
I would find it simpler eg. for such config:
```
project {
  modules = [a, b, c]
  name = x
}
```

to have just
```
val projectName = intelliJ.config[String]("project.name")
val projectName = intelliJ.config[Seq[String]]("project.modules")
```

instead of 

```
class Test extends ConfigFormat {
  case class TestProject(name: String, modules: Seq[String]
  import pureconfig.generic.auto._
  val expectedProject = intelliJ.config[TestProject]("project")
}
```

this is for the case when we have primitives in lists. I would suggest otherwise when we deal with lists of structures:
```
project {
  modules = [
    { name = a, contentRoots = [] },
    { name = b, contentRoots = [x/y] }
  ]
}
```

In such case it is worth to pull out the advanced apis to generate config and use it for easier comparison, etc. as it would be simpler than alternatives.

To sum up, I'd always go with what is more simple and straightforward.

driver/examples/src/test/resources/projects/cats/ideprobe.conf
```diff
@@ -0,0 +1,11 @@
+include required(classpath("ideprobe.conf"))
+
+probe {
+  workspace {
+    path = "https://github.com/typelevel/cats"
+  }
```
marek1840, line 6: this could be `workspace.path = ???`

driver/examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,47 @@
...
+    val project = intelliJ.probe.projectModel(projectRef)
+    val projectModules = project.modules.map(_.name)
+    val modulesFromConfig = intelliJ.config.get[Seq[String]]("test.modules").get
+
+    val missingModules = modulesFromConfig.diff(projectModules)
+    Assert.assertTrue(s"Modules $missingModules are missing", missingModules.isEmpty)
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = Array("projects/shapeless/ideprobe.conf", "projects/cats/ideprobe.conf"))
```
marek1840, line 34: no test for dokka?

build.sbt
```diff
@@ -101,6 +104,10 @@ lazy val junitDriver = module("junit-driver", "driver/bindings/junit")
 lazy val scalaTests = testModule("scala-tests", "extensions/scala/tests")
   .dependsOn(junitDriver)
 
+lazy val examples = testModule("examples", "driver/examples")
```
lukaszwawrzyk, line 107: I think we want it at root level (just "examples")

driver/examples/src/test/resources/ideprobe.conf
```diff
@@ -0,0 +1,15 @@
+probe {
+
+  driver {
+    headless = false
+    launch.timeout = "600 seconds"
```
lukaszwawrzyk, line 5: headless.false is default, launch timeout also should be ok with default. Generally this config is quite messy

driver/examples/src/test/resources/projects/cats/ideprobe.conf
lukaszwawrzyk: we accept empty workspace as default so this doesn't even have to be here.

driver/examples/src/test/resources/projects/dokka/ideprobe.conf
```diff
@@ -0,0 +1,9 @@
+include required(classpath("ideprobe.conf"))
+
+workspace {
+  path = "https://github.com/Kotlin/dokka"
+}
+
+test {
+  modules = ["ws"]
```
lukaszwawrzyk, line 8: can we list (in all files) more modules, preferably meaningful names that someone could think that come from the library mentioned in workspace path?

driver/examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  def fixtureFromConfig(configName: String): IntelliJFixture =
+    IntelliJFixture.fromConfig(Config.fromClasspath(configName))
+
+  @ParameterizedTest
+  @ValueSource(
+    strings = Array("projects/shapeless/ideprobe.conf", "projects/cats/ideprobe.conf", "projects/dokka/ideprobe.conf")
+  )
+  def verifyModulesPresent(configName: String): Unit = fixtureFromConfig(configName).run { intelliJ =>
+    intelliJ.probe.preconfigureJDK()
```
lukaszwawrzyk, line 23: it is not needed anymore for latest scala plugin nightly

driver/examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,47 @@
...
+    val projectRef = intelliJ.probe.openProject(intelliJ.workspace)
+    val project = intelliJ.probe.projectModel(projectRef)
+    val projectModules = project.modules.map(_.name)
+    val modulesFromConfig = intelliJ.config.get[Seq[String]]("test.modules").get
```
lukaszwawrzyk, line 27: just use config.apply `config[Seq[String]]("test.modules")`

driver/examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,47 @@
...
+
+    val missingModules = modulesFromConfig.diff(projectModules)
+    Assert.assertTrue(s"Modules $missingModules are missing", missingModules.isEmpty)
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = Array("projects/shapeless/ideprobe.conf", "projects/cats/ideprobe.conf"))
+  def runTestsInModules(configName: String): Unit = fixtureFromConfig(configName).run { intelliJ =>
+    intelliJ.probe.preconfigureJDK()
```
lukaszwawrzyk, line 36: also not needed

driver/examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,47 @@
...
+    val projectRef = intelliJ.probe.openProject(intelliJ.workspace)
+    val project = intelliJ.probe.projectModel(projectRef)
+    val modulesFromConfig = intelliJ.config.get[Seq[String]]("test.modules").get
+    val moduleRefs = modulesFromConfig.map(ModuleRef(_, project.name))
```
lukaszwawrzyk, line 40: First thing: if you have ProjectRef available, it is better to use it instead of project.name overload (just safer). 
Another thing is that we don't need to use project ref at all in single project scenario, as all methods accepting ProjectRef have default for ProjectRef.Default, which resolves to currently open project (if there is more than one it will fail). So you can simply use `ModuleRef(_)`

driver/examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,47 @@
...
+    val runConfigs = moduleRefs.map(JUnitRunConfiguration(_, None, None, None, None))
```
lukaszwawrzyk, line 41: `JUnitRunConfiguration` has factory methods in companion object like `JUnitRunConfiguration.module(_)`

driver/examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,47 @@
...
+
+    runConfigs.map(intelliJ.probe.run).foreach { result =>
```
lukaszwawrzyk, line 43: it would be ideal to first collect all failures and log them (so that user sees which test failed) and only then fail.

driver/examples/src/test/resources/projects/cats/ideprobe.conf
MichalBednarz: what doesn't need to be there? can somehow the path = ??? suffice?

build.sbt
```diff
@@ -99,7 +103,43 @@ lazy val junitDriver = module("junit-driver", "driver/bindings/junit")
...
+      intellijPluginName := "ideprobe-scala",
+      packageArtifactZipFilter := { file: File =>
+        // We want only this main jar to be packaged, all the library dependencies
+        // are already in the probePlugin which will be available in runtime as we
+        // depend on it in plugin.xml.
+        // The packaging plugin is created to support one plugin per build, so there
+        // seems to be no way to prevent including probePlugin.jar in the dist reasonable way.
+        file.getName == "scala-probe-plugin.jar"
+      },
+      intellijPlugins += "org.intellij.scala:2020.2.734".toPlugin,
```
lukaszwawrzyk, line 130: does it work finally? ;p

build.sbt
```diff
@@ -99,7 +103,43 @@ lazy val junitDriver = module("junit-driver", "driver/bindings/junit")
...
+      name := "scala-probe-plugin"
+    )
+
+lazy val scalaProbeDriver = project(id = "scala-probe-driver", path = "extensions/scala/driver", publish = true)
```
lukaszwawrzyk, line 134: for scala-probe ensure all modules that need to be published are (see the analogous modules for probe itself), for example we need to publish plugin itself. Also make sure that artifact naming is consistent with currently published artifacts. Also make sure that crossScalaVersions for scala-probe only has scala 2.12.

build.sbt
```diff
@@ -99,7 +103,43 @@ lazy val junitDriver = module("junit-driver", "driver/bindings/junit")
   )
 
 lazy val scalaTests = testModule("scala-tests", "extensions/scala/tests")
-  .dependsOn(junitDriver)
+  .dependsOn(junitDriver, scalaProbeDriver)
+  .usesIdeaPlugin(scalaProbePlugin)
+
+
+lazy val examples = testModule("examples", "examples")
```
lukaszwawrzyk, line 110: Would be nice to reorder projects in build.sbt according to how they depend on each other. 
probe modules
scala-probe modules
examples

and within probes: api, plugin, driver, tests

build.sbt
```diff
@@ -99,7 +103,43 @@ lazy val junitDriver = module("junit-driver", "driver/bindings/junit")
...
+      name := "scala-probe-plugin"
+    )
+
+lazy val scalaProbeDriver = project(id = "scala-probe-driver", path = "extensions/scala/driver", publish = true)
+  .enablePlugins(BuildInfoPlugin)
+  .disableIdeaPluginDevelopment
+  .dependsOn(scalaProbeApi, junitDriver)
+  .settings(
+    name := "scala-probe-driver",
+    buildInfoKeys := Seq[BuildInfoKey](version),
```
lukaszwawrzyk, line 140: no need to setup buildInfo, this class is already generated and setup for ideprobe that you depend on,.

examples/src/test/resources/projects/dokka/ideprobe.conf
```diff
@@ -0,0 +1,14 @@
+include required(classpath("ideprobe.conf"))
+
+probe {
+  workspace.path = "https://github.com/Kotlin/dokka"
+}
+
+test {
```
lukaszwawrzyk, line 7: what is this block for?

examples/src/test/resources/projects/shapeless/ideprobe.conf
```diff
@@ -0,0 +1,14 @@
+include required(classpath("ideprobe.conf"))
+
+probe {
+  workspace.path = "https://github.com/milessabin/shapeless"
+}
+
+test {
```
lukaszwawrzyk, line 7: seems like duplication

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
+import org.virtuslab.ideprobe.protocol.{BuildScope, JUnitRunConfiguration, ModuleRef, Setting}
+import org.virtuslab.intellij.scala.SbtTestSuite
+import org.virtuslab.intellij.scala.protocol.SbtProjectSettingsChangeRequest
+
+class ModuleTest extends SbtTestSuite {
+  @ParameterizedTest
+  @ValueSource(
+    strings = Array("projects/shapeless/ideprobe.conf", "projects/cats/ideprobe.conf", "projects/dokka/ideprobe.conf")
```
lukaszwawrzyk, line 15: probably nicer to read if each array element is in new line

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,53 @@
...
+  )
+  def verifyModulesPresent(configName: String): Unit = fixtureFromConfig(configName).run { intelliJ =>
+    deleteIdeaSettings(intelliJ)
+    intelliJ.probe.openProject(intelliJ.workspace)
+    val project = intelliJ.probe.projectModel()
+    val projectModules = project.modules.map(_.name)
```
lukaszwawrzyk, line 21: if you rebase I believe there is moduleNames method

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,53 @@
...
+    val modulesFromConfig = intelliJ.config[Seq[String]]("modules.verify")
+
+    val missingModules = modulesFromConfig.diff(projectModules)
+    Assert.assertTrue(s"Modules $missingModules are missing", missingModules.isEmpty)
+  }
+
+  @ParameterizedTest
+  @ValueSource(
+    strings = Array("projects/shapeless/ideprobe.conf", "projects/cats/ideprobe.conf", "projects/dokka/ideprobe.conf")
```
lukaszwawrzyk, line 30: if there is nothing in shapeless or cats or dokka directory it is porbably simpler to just have `projects/shapeless.conf` etc.

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,53 @@
...
+  )
+  def runTestsInModules(configName: String): Unit = fixtureFromConfig(configName).run { intelliJ =>
+    deleteIdeaSettings(intelliJ)
+    val projectRef = intelliJ.probe.openProject(intelliJ.workspace)
+    if(Files.exists(intelliJ.workspace.resolve("build.sbt"))) {
+      intelliJ.probe.setSbtProjectSettings(
+        SbtProjectSettingsChangeRequest(
+          useSbtShellForImport = Setting.Changed(true),
+          useSbtShellForBuild = Setting.Changed(true),
+          allowSbtVersionOverride = Setting.Changed(false)
```
lukaszwawrzyk, line 40: Do we want to explicitly disable it? why?

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,53 @@
...
+        )
+      )
+    }
+    val modulesFromConfig = intelliJ.config[Seq[String]]("modules.test")
+    val moduleRefs = modulesFromConfig.map(ModuleRef(_))
+    val runConfigs = moduleRefs.map(JUnitRunConfiguration.module)
+    val buildRes = intelliJ.probe.build(BuildScope.modules(projectRef, modulesFromConfig: _*))
```
lukaszwawrzyk, line 47: `bulld(...).assertSuccess()` or something, if you rebase ;p

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -0,0 +1,53 @@
...
+    Assert.assertTrue("Module build failed", !buildRes.hasErrors)
+
+    val result = runConfigs.map(config => config.module -> intelliJ.probe.run(config)).toMap
+    Assert.assertTrue(s"Tests in modules ${result.values} failed", result.values.forall(_.isSuccess))
```
lukaszwawrzyk, line 51: If only one module fails, the message won't be true.

extensions/scala/api/src/main/scala/org/virtuslab/intellij/scala/protocol/SbtEndpoints.scala
```diff
@@ -0,0 +1,14 @@
+package org.virtuslab.intellij.scala.protocol
+
+import org.virtuslab.ideprobe.jsonrpc.JsonRpc.Method.Request
+import org.virtuslab.ideprobe.protocol.ProjectRef
+import org.virtuslab.ideprobe.jsonrpc.PayloadJsonFormat._
+import pureconfig.generic.auto._
+
+object SbtEndpoints {
```
lukaszwawrzyk, line 8: I would replace naming in this PR from Sbt to Scala as it is scala-probe and sbt is just a part of it. We may want to add more endpoints that will be scala specific but not sbt specific.

extensions/scala/driver/src/main/scala/org/virtuslab/intellij/scala/SbtTestSuite.scala
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * The presence of .idea can prevent automatic import of gradle project
+   */
+  protected def deleteIdeaSettings(intelliJ: RunningIntelliJFixture) = {
+    val path = intelliJ.workspace.resolve(".idea")
+    if (Files.exists(path)) path.delete()
+  }
+
+  protected def fixtureFromConfig(configName: String): IntelliJFixture =
+    transformFixture(IntelliJFixture.fromConfig(Config.fromClasspath(configName)))
```
lukaszwawrzyk, line 19: don't implement it directly, just call on to another overoad of `fixtureFromConfig`. You can even add overload that accepts config *path* (not name, as you named it) as string, to the IntegrationTestSuite.

extensions/scala/driver/src/main/scala/org/virtuslab/intellij/scala/SbtTestSuite.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  val scalaProbePlugin: Plugin = InternalPlugins.bundle("ideprobe-scala")
+
+  override protected def transformFixture(fixture: IntelliJFixture): IntelliJFixture = {
+    fixture
+      .withPlugin(scalaProbePlugin)
+      .withAfterIntelliJInstall { (_, inteliJ) =>
+        inteliJ.paths.plugins.resolve("ideprobe/lib/scala-library.jar").delete()
```
lukaszwawrzyk, line 27: worth commenting it

extensions/scala/driver/src/main/scala/org/virtuslab/intellij/scala/SbtTestSuite.scala
```diff
@@ -0,0 +1,32 @@
...
+    val path = intelliJ.workspace.resolve(".idea")
+    if (Files.exists(path)) path.delete()
+  }
+
+  protected def fixtureFromConfig(configName: String): IntelliJFixture =
+    transformFixture(IntelliJFixture.fromConfig(Config.fromClasspath(configName)))
+
+  val scalaProbePlugin: Plugin = InternalPlugins.bundle("ideprobe-scala")
+
+  override protected def transformFixture(fixture: IntelliJFixture): IntelliJFixture = {
```
lukaszwawrzyk, line 23: add an issue to make new probes more pluggable (registering multiple transformers)

extensions/scala/probePlugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -0,0 +1,15 @@
+<idea-plugin>
+    <id>org.virtuslab.ideprobe.scalaplugin</id>
+    <name>Virtuslab IdeProbe Scala</name>
+    <vendor email="info@virtuslab.com" url="http://www.virtuslab.com">VirtusLab</vendor>
+
+    <depends>com.intellij.modules.lang</depends>
+    <depends>com.intellij.modules.java</depends>
+    <depends>org.virtuslab.ideprobe</depends>
+    <depends>org.intellij.scala</depends>
+    <idea-version since-build="191.*" until-build="201.*"/>
```
lukaszwawrzyk, line 10: we want to handle 202 as well actually, so best remove until at all or bump it to 202

extensions/scala/probePlugin/src/main/scala/org/virtuslab/intellij/scala/probe/SbtProbeHandlerContributor.scala
```diff
@@ -0,0 +1,14 @@
+package org.virtuslab.intellij.scala.probe
```
lukaszwawrzyk, line 1: let's use org.virtuslab.ideprobe.scala package for this extension.

extensions/scala/probePlugin/src/main/scala/org/virtuslab/intellij/scala/probe/handlers/SbtSettings.scala
```diff
@@ -0,0 +1,38 @@
...
+      def setSetting[A](setting: Setting[A])(f: (SbtProjectSettingsFromPlugin, A) => Unit): Unit = {
+        setting.foreach(value => f(sbtSettings, value))
+      }
+
+      setSetting(toSet.useSbtShellForImport)(_.setUseSbtShellForImport(_))
+      setSetting(toSet.useSbtShellForBuild)(_.setUseSbtShellForBuild(_))
+      setSetting(toSet.allowSbtVersionOverride)(_.setAllowSbtVersionOverride(_))
+    }
+
+  private def getSbtSettings(project: Project) = SbtProjectSettingsFromPlugin.forProject(project).get
```
lukaszwawrzyk, line 37: add extends IntelliJApi for the object and use `getOrElse(error("No settings (...) probably not sbt project (...)"))

extensions/scala/tests/src/test/scala/org/virtuslab/intellij/extensions/SbtSettingsTest.scala
```diff
@@ -0,0 +1,29 @@
+package org.virtuslab.intellij.extensions
```
lukaszwawrzyk, line 1: also fix package here, basically everywhere.

extensions/scala/tests/src/test/scala/org/virtuslab/intellij/extensions/SbtSettingsTest.scala
```diff
@@ -0,0 +1,29 @@
...
+      SbtProjectSettingsChangeRequest(
+        useSbtShellForImport = Setting.Changed(true),
+        useSbtShellForBuild = Setting.Changed(true),
+        allowSbtVersionOverride = Setting.Changed(false)
+      )
+    )
+    val expectedSettings = SbtProjectSettings(useSbtShellForImport = true, useSbtShellForBuild = true, allowSbtVersionOverride = false)
+    val actualSettings = intelliJ.probe.getSbtProjectSettings()
+
+    Assert.assertTrue("Settings are not changed correctly", actualSettings == expectedSettings)
```
lukaszwawrzyk, line 26: use assertEquals for clear error message

extensions/scala/driver/src/main/scala/org/virtuslab/intellij/scala/SbtTestSuite.scala
lukaszwawrzyk: also optional dependency on scala plugin in plugin.xml is worth documenting. One place should contain explanation (probably in code) and the other should point to the first one

build.sbt
MichalBednarz: Yes it does, somehow

build.sbt
MichalBednarz: done

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
MichalBednarz: done

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
MichalBednarz: done

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
MichalBednarz: done

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
MichalBednarz: done

extensions/scala/api/src/main/scala/org/virtuslab/intellij/scala/protocol/SbtEndpoints.scala
MichalBednarz: done

extensions/scala/driver/src/main/scala/org/virtuslab/intellij/scala/SbtTestSuite.scala
MichalBednarz: done

extensions/scala/driver/src/main/scala/org/virtuslab/intellij/scala/SbtTestSuite.scala
MichalBednarz: done

extensions/scala/driver/src/main/scala/org/virtuslab/intellij/scala/SbtTestSuite.scala
MichalBednarz: done

extensions/scala/probePlugin/src/main/resources/META-INF/plugin.xml
MichalBednarz: done

extensions/scala/probePlugin/src/main/scala/org/virtuslab/intellij/scala/probe/SbtProbeHandlerContributor.scala
MichalBednarz: done

extensions/scala/probePlugin/src/main/scala/org/virtuslab/intellij/scala/probe/handlers/SbtSettings.scala
MichalBednarz: done

extensions/scala/tests/src/test/scala/org/virtuslab/intellij/extensions/SbtSettingsTest.scala
MichalBednarz: done

extensions/scala/tests/src/test/scala/org/virtuslab/intellij/extensions/SbtSettingsTest.scala
MichalBednarz: done

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
MichalBednarz: `checkBuildPanelErrors()` ?

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
MichalBednarz: We can leave it unchanged.

examples/src/test/resources/projects/shapeless/ideprobe.conf
MichalBednarz: done

examples/src/test/resources/projects/dokka/ideprobe.conf
MichalBednarz: unneeded, done

build.sbt
MichalBednarz: done

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
lukaszwawrzyk: no, the BuildResult should have a method to assert that it passed

build.sbt
MichalBednarz: seems to be ok now

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
MichalBednarz: ok done

build.sbt
```diff
@@ -98,8 +102,43 @@ lazy val junitDriver = module("junit-driver", "driver/bindings/junit")
...
   )
 
+lazy val scalaProbeApi = project(id = "scala-probe-api", path = "extensions/scala/api", publish = true).dependsOn(api)
+
+lazy val scalaProbePlugin =
+  ideaPluginModule(id = "scala-probe-plugin", path = "extensions/scala/probePlugin", publish = true)
+    .dependsOn(probePlugin, scalaProbeApi)
+    .settings(
+      intellijPluginName := "ideprobe-scala",
+      crossScalaVersions := List(scala212),
```
lukaszwawrzyk, line 112: I think we probably need to only publish 2.12 for all of scala-probe modules. It will be difficult to run driver on 2.13 and internals on 2.12

extensions/scala/driver/src/main/scala/org/virtuslab/ideprobe/scala/ScalaTestSuite.scala
```diff
@@ -0,0 +1,33 @@
...
+      .withAfterIntelliJInstall { (_, inteliJ) =>
+        // The scala-library from ideprobe plugin causes conflict with the scala-library from
+        // scala plugin. This is why we delete one of them. We declare the scala-library as an
+        // optional dependency with config file probePlugin/src/main/resources/META-INF/scala-plugin.xml
+        // so that ideprobe plugin can be loaded regardless of the missing scala-library.
+        inteliJ.paths.plugins.resolve("ideprobe/lib/scala-library.jar").delete()
+      }
+  }
+
+  implicit def sbtProbeDriver(driver: ProbeDriver): ScalaProbeDriver = ScalaProbeDriver(driver)
```
lukaszwawrzyk, line 32: missed that rename ;p

extensions/scala/probePlugin/src/main/scala/org/virtuslab/ideprobe/scala/handlers/SbtSettings.scala
```diff
@@ -0,0 +1,39 @@
...
+
+      setSetting(toSet.useSbtShellForImport)(_.setUseSbtShellForImport(_))
+      setSetting(toSet.useSbtShellForBuild)(_.setUseSbtShellForBuild(_))
+      setSetting(toSet.allowSbtVersionOverride)(_.setAllowSbtVersionOverride(_))
+    }
+
+  private def getSbtSettings(project: Project) =
+    SbtProjectSettingsFromPlugin
+      .forProject(project)
+      .getOrElse(error(s"No settings for ${project.getName}, probably not sbt project."))
```
lukaszwawrzyk, line 38: `not an sbt project`

extensions/scala/driver/src/main/scala/org/virtuslab/ideprobe/scala/ScalaTestSuite.scala
MichalBednarz: done

extensions/scala/probePlugin/src/main/scala/org/virtuslab/ideprobe/scala/handlers/SbtSettings.scala
MichalBednarz: done

build.sbt
```diff
@@ -73,10 +79,16 @@ lazy val driver = module("driver", "driver/sources")
     libraryDependencies += Dependencies.nuProcess,
     libraryDependencies += Dependencies.remoteRobot,
     libraryDependencies += Dependencies.remoteRobotFixtures,
-    buildInfoKeys := Seq[BuildInfoKey](version, intellijBuild, "robotVersion" -> Dependencies.remoteRobot.revision),
+    buildInfoKeys := Seq[BuildInfoKey](
+      version,
+      BuildInfoKey.map(intellijBuild) { case (k, v) => k -> (v + "-EAP-SNAPSHOT") },
```
marek1840, line 84: It will fail if the version is like `2020.1`. Can we reuse the logic from the plugin itself here? They are appending the suffix only to the EAPs right?

driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
```diff
@@ -21,6 +21,18 @@ object Resource extends ConfigFormat {
     cur.asString.map(Resource.from)
   }
 
+  def exists(uri: URI): Boolean = from(uri) match {
+    case Jar(uri) => true // TODO
```
lukaszwawrzyk, line 25: Files.exists might also work here

build.sbt
```diff
@@ -1,12 +1,15 @@
...
 val scala213 = "2.13.1"
 
 skip in publish := true
 
-scalaVersion.in(ThisBuild) := scala213
-intellijBuild.in(ThisBuild) := "202.6397.20-EAP-SNAPSHOT"
+scalaVersion.in(ThisBuild) := scala212
+// -EAP-SNAPSHOT suffix results in incorrect url for the intellij scala plugin
+// it is appended automatically for intellij sdk dependency url and manually for the BuildInfo
+intellijBuild.in(ThisBuild) := "202.6397.20"
```
lukaszwawrzyk, line 12: we could just bump to stable 2020.2.1 (still needs a number like 202.xxxx.xx)

build.sbt
```diff
@@ -74,32 +81,63 @@ lazy val driver = module("driver", "driver/sources")
...
+
 lazy val scalaTests = testModule("scala-tests", "extensions/scala/tests")
-  .dependsOn(junitDriver)
+  .dependsOn(junitDriver, scalaProbeDriver)
+  .usesIdeaPlugin(scalaProbePlugin)
+
+
+lazy val examples = testModule("examples", "examples")
+  .dependsOn(junitDriver, scalaProbeDriver)
+  .usesIdeaPlugin(scalaProbePlugin)
```
lukaszwawrzyk, line 139: this shouldn't be necessary

MichalBednarz: the scala probe plugin is not available on examples' classpath without that

build.sbt
MichalBednarz: done

build.sbt
lukaszwawrzyk: this means that something is wrong, it must be available. Imagine someone uses this from different project. Testing should just work by depending on scalaProbeDriver and driver. Take a look why it works for base ideprobe.

Issue #20, opened 2020-07-31
Resolve duplication regarding Git setups


lukaszwawrzyk:
We have at least 3 places that refer to git that I recall:
- Setting workspace from git
- Fetching pants from git (in pants plugin repo)
- Fetching a plugin from git sources to build it

Some of these only allow to specify uri and optionally branch, some also allow for tag or commit. They all are used from .conf files. I think all of them are part of broader ADT.

It would be really useful to have just one `GitReference` class/trait with variants to specify tag, branch, ref or anything relevant for fetching some revision of source, and also to have a robust method to clone it, cache it and setup in some place (as we have it implemented separately for each of these cases).

This refactoring will not only cleanup the code, enable commit cloning in case of building plugins from sources but also make it easy to setup custom configuratble actions in fixtures that use git (like setting up pants from sources).

PR #28, opened 2020-08-04
Enable scala tests from scala plugin master branch

lukaszwawrzyk:


driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/ResourceProvider.scala
```diff
@@ -32,14 +32,17 @@ object ResourceProvider {
...
-          .createTempFile("cached-resource", "-tmp")
-          .append(stream())
-          .moveTo(cachedResource)
+        val stream = openStream()
+        try {
+          Files
+            .createTempFile("cached-resource", "-tmp")
+            .append(stream)
+            .moveTo(cachedResource)
+        } finally stream.close()
```
marek1840, line 45: `append` extension should already close the stream

driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/SourceRepository.scala
```diff
@@ -7,7 +7,7 @@ import pureconfig.generic.semiauto.deriveReader
 
 sealed trait SourceRepository
 object SourceRepository extends ConfigFormat {
-  final case class Git(path: URI, branch: Option[String]) extends SourceRepository
+  final case class Git(path: URI, ref: Option[String]) extends SourceRepository
```
marek1840, line 10: Reader should probably accept any of those: `ref`, `tag`, `branch`, `commit` ?

extensions/scala/tests/src/main/scala/org/virtuslab/intellij/extensions/ScalaPluginBuilder.scala
```diff
@@ -39,26 +39,25 @@ object ScalaPluginBuilder extends DependencyBuilder(Id("scala")) {
...
     val localRepo = Files.createTempDirectory("scala-plugin-repo")
-    val command: Array[String] = repository.branch match {
-      case None         => Array("git", "clone", repository.path.toString, localRepo.toString)
-      case Some(branch) => Array("git", "clone", repository.path.toString, branch, localRepo.toString)
+    val cloned = Shell.run("git", "clone", repository.path.toString, localRepo.toString)
+    if (cloned.exitCode != 0) throw new IllegalStateException(s"Could not clone git $repository")
+    repository.ref.foreach { ref =>
+      val checkout = Shell.run(in = localRepo, "git", "checkout", ref)
+      if (checkout.exitCode != 0) throw new IllegalStateException(s"Could not checkout $ref in $repository")
     }
```
marek1840, line 47: I'm not sure if that is any simpler. Is it because we can clone only branches?

driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/ResourceProvider.scala
lukaszwawrzyk: Ah, ok. I will revert. Though a bit confusing. I guess the same function that opens stream should close it.

driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/SourceRepository.scala
lukaszwawrzyk: For now I just enabled the use of this for all ref scenarios. There is an issue https://github.com/VirtusLab/ide-probe/issues/20 that should make sure that all git related places have common model and methods.

extensions/scala/tests/src/main/scala/org/virtuslab/intellij/extensions/ScalaPluginBuilder.scala
lukaszwawrzyk: It is not meant to be simpler, it is meant to support checking out commits. We can only clone branches, yes, so it would not work with commit hashes.

PR #30, opened 2020-08-05
Add timeout for awaiting notification

lukaszwawrzyk:


driver/tests/src/test/scala/org/virtuslab/ideprobe/ProbeDriverTest.scala
```diff
@@ -250,12 +251,15 @@ final class ProbeDriverTest extends IntegrationTestSuite with Assertions {
...
   }
 
+  @tailrec
   private def retry[A](times: Int)(action: => A): A = {
     try {
       action
     } catch {
       case e: WaitForConditionTimeoutException =>
         if (times > 0) {
+          println("Failed to find element, retrying...")
```
marek1840, line 261: isn't the action generic? We might not always be looking for an element

lukaszwawrzyk: this exception will only be thrown from `find`. Anyway I just added it for debugging purposes as it failed again on this PR. Eventually, when we know what is wrong we can make this generic.

PR #38, opened 2020-08-21
Allow reusing existing workspace

lukaszwawrzyk:
Useful if we don't care about multiple tests interacting with the same workspace (e.g. if each test cleanups after itself or doesn't write anything that can affect other tests). Also useful when setting up repo is costly and/or custom, so it can be prepared externally.


tpasternak (approved):
LGTM

PR #39, opened 2020-08-21
More robust way to delete directories

odisseus:
Something strange happens while the directory tree is traversed.

### DIFF
--- a/api/src/main/scala/org/virtuslab/ideprobe/ProbeExtensions.scala
+++ b/api/src/main/scala/org/virtuslab/ideprobe/ProbeExtensions.scala
@@ -9,12 +9,13 @@ import java.net.URI
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.FileVisitResult
-import java.nio.file.NoSuchFileException
 import java.nio.file.StandardCopyOption
 import java.nio.file.StandardOpenOption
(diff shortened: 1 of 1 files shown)


odisseus:
Is printing to stdout the accepted way for logging?

api/src/main/scala/org/virtuslab/ideprobe/ProbeExtensions.scala
```diff
@@ -173,3 +170,37 @@ trait ProbeExtensions {
     }
   }
 }
+
+object ProbeExtensions {
+  private class DeletingVisitor(root: Path) extends SimpleFileVisitor[Path] {
+    override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
+      if(!Files.isDirectory(file))  Files.delete(file)
```
lukaszwawrzyk, line 177: You should be able to check if this is directory by calling a method on attrs. That would save us an IO call.

lukaszwawrzyk (approved):

api/src/main/scala/org/virtuslab/ideprobe/ProbeExtensions.scala
odisseus: Done

Issue #22, opened 2020-07-31
IntelliJ instance should be deleted after failed test  


MichalBednarz:
Each test run creates a new copy of IntelliJ instance. For now, the instance stored in `/tmp/` isn't cleaned up after the failing test. It results in a memory overflow caused by limited space in the `/tmp/` which can be resolved only by manually deleting instances or by the reboot. 

In addition, we can remove `ideprobe-workspace*` directories cause they are left empty after the test execution.

odisseus:
I'm not sure it's a good idea to clean the environment after a failed test. The user might want to inspect it in order to find the cause of failure.

marek1840:
> I'm not sure it's a good idea to clean the environment after a failed test. The user might want to inspect it in order to find the cause of failure.

I agree -  it makes sense. Would you, @odisseus, be willing to change the logic so that we remove the workspace and ide only after a successful or interrupted test?

Also what about CI? We don't want to hog the already limited space there. Additionally, people are potentially interested in all the failures we can get - would it make sense to delete these directories on failure only on the CI?

odisseus:
The argument for keeping the environment of a failed test holds for the CI as well. I think it is best to keep these environments unless you are very constrained on space. Ideally, the environment of a failed test would be archived and made available for download and inspection.

I'll try to implement the cleanup logic anyway, and then we can decide when it should be invoked.

odisseus:
Files and directories that still remain after a successful test run:
```
scala-plugin-repo3777509081734216570
ideprobe
liblz4-java4000184621870245838.so
bsp-launcher1787856568650195900
jna-1309000474
hsperfdata_virtuslab
```

lukaszwawrzyk:
Oh, this is good to know. Most of these are out of our control, I think it is ok to leave them as they are. `ideprobe` is expected to stay, we keep cache and outputs there. 
As for `scala-plugin-repo3777509081734216570` it comes from `ScalaPluginBuilder`. It now just clones repo to temp dir, builds it there and returns the built file that lives there that I think is put in cache using ResourceProvider. We should somehow reorganize code there, so that we can delete the directory after the disted plugin is cached.
Probably you will need to do exactly the same thing for pants plugin, so when doing that you may want to also refactor this one.

PR #40, opened 2020-08-23
Delete the empty parent directory after the workspace is deleted

odisseus:
This pull request reduces the clutter in the `/tmp` directory by deleting the leftover empty base directories.

Since the base directory is created along with the workspace directory, I believe it is appropriate to delete it along with the workspace too.

This pull request is related to the issue #22.

### DIFF
--- a/driver/bindings/junit/src/main/scala/org/virtuslab/ideprobe/WorkspaceProvider.scala
+++ b/driver/bindings/junit/src/main/scala/org/virtuslab/ideprobe/WorkspaceProvider.scala
@@ -63,7 +63,10 @@ sealed trait WorkspaceTemplate extends WorkspaceProvider {
     workspace
   }
 
-  override def cleanup(path: Path): Unit = path.delete()
+  override def cleanup(path: Path): Unit = {
+    path.delete()
+    path.getParent.delete()
+  }
 
   def setupIn(workspace: Path): Unit
 }


driver/bindings/junit/src/main/scala/org/virtuslab/ideprobe/IntelliJFixture.scala
```diff
@@ -48,6 +48,8 @@ final case class IntelliJFixture(
 
   def deleteWorkspace(workspace: Path): Unit = {
     workspace.delete()
+    val workspaceBase = workspace.getParent
+    if (Files.isDirectory(workspaceBase) && workspaceBase.toFile.list().isEmpty) Files.delete(workspaceBase)
```
marek1840, line 52: or you could just do: `workspace.getParent.delete()` instead of `workspace.delete()`. It should still be a directory controlled by the probe, so it should be safe to delete

odisseus: My intent was to make it clear that the parent directory is assumed to be empty. If the directory is not empty (which seems to happen in some tests), that's an unexpected situation, and I'd rather leave it as it is than risk deleting important data.

lukaszwawrzyk: I refactored workspace setup a bit. Now WorkspaceTemplate contains the cleanup logic. I believe Marek is right. If we setup directory in directory as a workspace, we should also clean it up, nothing strange in this logic. 
In fact for fastpass tests, the `ws.parent` will not be empty, as fastpass will setup <ws>/bsp-projects along with <ws>/ws, so it is desirable to have it deleted together.

lukaszwawrzyk (approved):

PR #43, opened 2020-09-02
Add scalatest support

MichalBednarz:


build.sbt
```diff
@@ -87,15 +87,16 @@ lazy val driver = module("driver", "driver/sources")
...
 lazy val driverTests = testModule("driver-tests", "driver/tests")
   .dependsOn(driver, junitDriver, api % "compile->compile;test->test")
   .usesIdeaPlugin(driverTestPlugin)
 
 lazy val probePlugin = ideaPluginModule("probe-plugin", "probePlugin", publish = true)
   .dependsOn(api)
-  .settings(intellijPluginName := "ideprobe")
+  .settings(
+    intellijPluginName := "ideprobe",
+    intellijPlugins += Dependencies.intellijScala.toPlugin
```
lukaszwawrzyk, line 98: This looks bad. This is what we have scala extension for...

lukaszwawrzyk (changes requested):
Generally all the changes for running scalatest should live in extensions/scala. 

Also why don't we use shapeless or cats anymore? They are much more popular.

Let's say this is ok temporarly, but as I said, we need to hook in to extension points that produce run configration based on context and possibly provide api to select one of existing run configuration types. This will make probe and the tests truly extensible.

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
```diff
@@ -1,35 +1,39 @@
...
 import com.intellij.compiler.options.CompileStepBeforeRun.MakeBeforeRunTask
-import com.intellij.execution.ExecutionManager
-import com.intellij.execution.RunManager
-import com.intellij.execution.RunnerAndConfigurationSettings
+import com.intellij.execution._
 import com.intellij.execution.application.ApplicationConfiguration
 import com.intellij.execution.executors.DefaultRunExecutor
-import com.intellij.execution.impl.RunManagerImpl
-import com.intellij.execution.impl.RunnerAndConfigurationSettingsImpl
+import com.intellij.execution.impl.{RunManagerImpl, RunnerAndConfigurationSettingsImpl}
```
odisseus, line 10: Do you have any conventions on the formatting of imports?

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
```diff
@@ -1,35 +1,39 @@
...
-import org.virtuslab.ideprobe.protocol.TestsRunResult
+import org.virtuslab.ideprobe.protocol.{
+  ApplicationRunConfiguration,
+  JUnitRunConfiguration,
+  ProcessResult,
+  TestRun,
+  TestStatus,
+  TestSuite,
+  TestsRunResult,
+  _
```
odisseus, line 33: Importing a bunch of single classes is not necessary when you also import the entire package with a wildcard.

api/src/main/scala/org/virtuslab/ideprobe/protocol/ApplicationRunConfiguration.scala
```diff
@@ -13,6 +13,12 @@ final case class JUnitRunConfiguration(
     directory: Option[String]
 )
 
+final case class ScalaTestRunConfiguration(module: ModuleRef,
+                                           packageName: Option[String],
+                                           className: Option[String],
+                                           testName: Option[String],
```
odisseus, line 19: It seems that the three optional fields in this class are intended to be mutually exclusive. In that case, a sum type (a sealed trait with extending case classes) would be a much better fit.
```scala
sealed abstract class ScalaTestRunConfiguration(val moduleRef: ModuleRef)

case class PackageConfiguration(override val moduleRef: ModuleRef, packageName: String) extends ScalaTestRunConfiguration(moduleRef)
case class ClassConfiguration(override val moduleRef: ModuleRef, className: String) extends ScalaTestRunConfiguration(moduleRef)
case class TestConfiguration(override val moduleRef: ModuleRef, testName: String) extends ScalaTestRunConfiguration(moduleRef)
```

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
lukaszwawrzyk: Perhaps we should setup scalafmt, as these import changes are distracting

api/src/main/scala/org/virtuslab/ideprobe/protocol/ApplicationRunConfiguration.scala
lukaszwawrzyk: I think TestConfiguration should have both `className` and `testName`. Initially I took this approach for junit when we had a different json serializer without good support of this. It would be generally better for the code to use ADT. I am hoping we can do this refactor along with a correct implementaton of running tests (hooking in to run configuration producers), so that it can be do once.

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -2,16 +2,56 @@ package org.virtuslab.ideprobe
...
+      JUnitRunConfiguration.method(moduleRef, className, methodName)
+    )
+
+    runConfigurations.map(intelliJ.probe.run).foreach { result =>
+      Assert.assertTrue(s"Test result $result should not be empty", result.suites.nonEmpty)
+    }
+  }
+
+  @Test
+  def runScalaTestTestsInDifferentScopes: Unit = fixtureFromConfig("projects/io.conf").run { intelliJ =>
```
lukaszwawrzyk, line 36: does ci run this?

lukaszwawrzyk (approved):

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
MichalBednarz: no, do we even want examples to be run on ci? They depend on repositories that can change in the future.

PR #49, opened 2020-09-14
Support class and method test run

MichalBednarz:


probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
```diff
@@ -36,54 +37,81 @@ object RunConfigurations extends IntelliJApi {
...
             )
           )
       }
       val selectedConfiguration = producer.getConfigurationSettings
       val transformedConfiguration = RunConfigurationTransformer.transform(selectedConfiguration)
+      transformedConfiguration.getConfiguration match {
+        case config: JUnitConfiguration => {
+          // sets the "use classpath of module:" configuration setting
+          // because it isn't set out of the box in JUnitConfiguration
+          config.getTestObject.getConfiguration.getConfigurationModule.setModule(module)
```
lukaszwawrzyk, line 101: I think there is a direct setModule method on JUnitConfiguration

MichalBednarz: It could have been just some local issue because I can't reproduce it anymore

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
```diff
@@ -36,44 +38,65 @@ object RunConfigurations extends IntelliJApi {
       dataContext.put(CommonDataKeys.PROJECT, project)
       dataContext.put(LangDataKeys.MODULE, module)
 
-      val projectPath = Paths.get(project.getBasePath)
-
-      // TODO getContentRoots returns an array, is taking head only reliable?
-      val moduleVirtualFile = ModuleRootManager.getInstance(module).getContentRoots.head
-      val moduleDir = runOnUISync {
-        PsiManager.getInstance(project).findDirectory(moduleVirtualFile)
+      val psiElement: PsiElement = (runConfiguration.packageName, runConfiguration.className, runConfiguration.methodName) match {
```
lukaszwawrzyk, line 41: what about directory?

lukaszwawrzyk (approved):

driver/sources/src/main/scala/org/virtuslab/ideprobe/ProbeDriver.scala
```diff
@@ -189,10 +189,12 @@ class ProbeDriver(
   def freezes: Seq[Freeze] = send(Endpoints.Freezes)
 
   /**
-   * Runs the specified test configuration with a test runner containing provided string
-   * or with the first available runner when the string is not present
+   * Runs the specified test configuration with a test runner containing provided runnerNameFragment
+   * or with the first available runner when the runnerNameFragment is not present
    */
-  def run(runConfiguration: TestRunConfiguration): TestsRunResult = send(Endpoints.RunTest, runConfiguration)
+  def run(runConfiguration: TestRunConfiguration, runnerNameFragment: Option[String]): TestsRunResult = {
```
lukaszwawrzyk, line 195: it would be better to make 2 overloads with runnerNameFragment: String and without, this method can be private

driver/tests/src/test/scala/org/virtuslab/ideprobe/ProbeDriverTest.scala
```diff
@@ -273,6 +264,41 @@ final class ProbeDriverTest extends IdeProbeFixture with Assertions with RobotPl
...
+      val moduleRef = ModuleRef("foo.test")
+
+      val moduleRunConfiguration = TestRunConfiguration.Module(moduleRef)
+      val runResult = intelliJ.probe.run(moduleRunConfiguration, None)
+      assert(runResult.suites.size == 2)
+
+      val directoryName = "java"
+      val directoryRunConfiguration = TestRunConfiguration.Directory(moduleRef, directoryName)
+      val directoryRunResult = intelliJ.probe.run(directoryRunConfiguration, None)
+      assert(directoryRunResult.suites.size == 2)
```
lukaszwawrzyk, line 280: I don't think this assert is macro based. We should use assertEquals(expected, actual) for better error messages

examples/src/test/resources/ideprobe.conf
```diff
@@ -3,6 +3,6 @@ probe {
     plugins = [
       { id = "org.intellij.scala", version = "2020.2.730", channel = "nightly" }
     ]
-    version = "202.6397.59-EAP-SNAPSHOT"
+    version = { build = "202.6397.59-EAP-SNAPSHOT" }
```
lukaszwawrzyk, line 6: better to use version.build = ...

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
```diff
@@ -51,8 +77,8 @@ class ModuleTest extends IdeProbeFixture with ScalaPluginExtension with RobotPlu
     val modulesFromConfig = intelliJ.config[Seq[String]]("modules.test")
     val runnerNameFragmentOpt = intelliJ.config.get[String]("runner")
     val moduleRefs = modulesFromConfig.map(ModuleRef(_))
-    val runConfigs = moduleRefs.map(moduleRef => TestRunConfiguration(moduleRef, runnerNameFragmentOpt))
-    runConfigs.map(config => config.module -> intelliJ.probe.run(config)).foreach {
+    val runConfigs = moduleRefs.map(moduleRef => TestRunConfiguration.Module(moduleRef))
```
lukaszwawrzyk, line 80: Indeed JUnitRunConfiguration was mapped to concrete run configuration directly, I think now TestScope is more accurate name for this class.

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
```diff
@@ -27,53 +30,79 @@ import scala.collection.convert.ImplicitConversions.`collection AsScalaIterable`
 import scala.concurrent.ExecutionContext
 
 object RunConfigurations extends IntelliJApi {
-  def execute(runConfiguration: TestRunConfiguration)(implicit ec: ExecutionContext): TestsRunResult =
+  def execute(testRunnerMatch: TestRunConfigurationMatch)(implicit ec: ExecutionContext): TestsRunResult =
     BackgroundTasks.withAwaitNone {
+      val TestRunConfigurationMatch(runConfiguration, runnerNameFragment) = testRunnerMatch
       val project = Projects.resolve(ProjectRef.Default)
```
lukaszwawrzyk, line 36: you should resolve project from module ref (just module.getProject)

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
```diff
@@ -27,53 +30,79 @@ import scala.collection.convert.ImplicitConversions.`collection AsScalaIterable`
...
+          val moduleVirtualFile = ModuleRootManager.getInstance(module).getContentRoots.head
+          val psiDirectory = read {
+            PsiManager.getInstance(project).findDirectory(moduleVirtualFile)
+          }
+          Option(psiDirectory).getOrElse(error(s"Directory of module ${module.getName} not found"))
+        }
+        case TestRunConfiguration.Directory(_, directoryName) => {
+          val moduleContentRoots = ModuleRootManager.getInstance(module).getContentRoots.head
+          val dirPath = Paths.get(moduleContentRoots.getPath, directoryName.replace(".", "/"))
+          val dirVirtualFile = VirtualFileManager.getInstance().findFileByNioPath(dirPath)
```
lukaszwawrzyk, line 54: I think we have VFS. util method for that

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
```diff
@@ -27,53 +30,79 @@ import scala.collection.convert.ImplicitConversions.`collection AsScalaIterable`
 import scala.concurrent.ExecutionContext
 
 object RunConfigurations extends IntelliJApi {
```
lukaszwawrzyk, line 32: Does this object also has runner for apps? If so it is ok, if not, maybe rename it for TestRunConfigurations? Anyway, current method names are quite confusing. Both named execute with very different behavior. Parameter type doesn't explain anymore what it is about. We should have something like `executeJUnit` and `executeGenerated`. Also I believe we don't need to have the `TestRunConfigurationMatch` case class, at least not here, I'd unpack the arguments.
I think such api would be much nicer:
```
object TestRunners {
  def runJUnit(scope: TestScope): TestRunResult = ???
  def runGenerated(scope: TestScope, runnerToSelect: Option[String]): TestRunResult = ???
}
```

lukaszwawrzyk (changes requested):

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
MichalBednarz: IMO the problem is that now TestScope took over the JUnitRunConfiguration (API for JUnit and generic test run became unified). We would have to run JUnitRunConfiguration with the object called TestScope then.

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
MichalBednarz: Actually, it has `def execute(runConfiguration: ApplicationRunConfiguration)`

lukaszwawrzyk: ok, so see my edit

examples/src/test/scala/org/virtuslab/ideprobe/ModuleTest.scala
lukaszwawrzyk: it is ok, we run JUnit based on TestScope. If we were to have some custom junit stuff, we can introduce case class JUnitRunConfiguration(scope: TestScope, otherParam: Type)

PR #53, opened 2020-09-24
Extract robot to separate extension

lukaszwawrzyk:


extensions/robot/driver/src/main/scala/org/virtuslab/ideprobe/robot/RobotProbeDriver.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  def closeTipOfTheDay(): Unit = {
+    Try(robot.mainWindow.find(query.dialog("Tip of the Day")).button("Close").click())
+  }
+
+  def checkBuildPanelErrors(): Unit = {
+    robot.findOpt(query.className("MultipleBuildsPanel")).foreach { buildPanel =>
+      val tree = buildPanel.find(query.className("Tree"))
+      val treeTexts = tree.fullTexts
```
tpasternak, line 60: Is this null-safe? (moved from other file, so ok to ignore)

extensions/robot/driver/src/main/scala/org/virtuslab/ideprobe/robot/RobotProbeDriver.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  def openProject(path: Path): ProjectRef = {
+    val ref = driver.openProject(path)
+    closeTipOfTheDay()
+    checkBuildPanelErrors()
+    ref
+  }
+
+  def closeTipOfTheDay(): Unit = {
+    Try(robot.mainWindow.find(query.dialog("Tip of the Day")).button("Close").click())
```
tpasternak, line 54: Is it intentional to catch everything here? ( know it was moved, so feel free to ignore)

tpasternak (approved):

extensions/robot/driver/src/main/scala/org/virtuslab/ideprobe/robot/RobotProbeDriver.scala
lukaszwawrzyk: I is not needed for anything, I just close it because it is annoying and can cover something that would be important for screenshot, I don't want to fail the test if flaky robot fails.

extensions/robot/driver/src/main/scala/org/virtuslab/ideprobe/robot/RobotProbeDriver.scala
lukaszwawrzyk: Find is null safe, it throws if it fails to find within timeout

PR #56, opened 2020-10-06
2020.3

lukaszwawrzyk:


ci/tests/2.13.1/test-probe
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+export IDEPROBE_DISPLAY=xvfb
+export JAVA_HOME=/usr/lib/jvm/java-11-openjdk
+sbt "scala-tests/testOnly org.virtuslab.ideprobe.scala.BspImportTest"
```
tpasternak, line 5: Why we are doing this separately?

lukaszwawrzyk: this is just leftover from my debugging, CI regenerates this script anyway. Thanks for finding

PR #65, opened 2020-10-16
Add configurable await idle

MichalBednarz:


examples/src/test/resources/ideprobe.conf
```diff
@@ -4,4 +4,9 @@ probe {
       { id = "org.intellij.scala", version = "2020.2.804", channel = "nightly" }
     ]
   }
+  endpoints.awaitIdle {
+    initialWait = "0 seconds"
+    newTaskWait = "1 seconds"
```
lukaszwawrzyk, line 9: "1 second" should work and is more english :p

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/BackgroundTasks.scala
```diff
@@ -3,9 +3,11 @@ package org.virtuslab.ideprobe.handlers
 import java.lang.reflect.Method
 import com.intellij.openapi.progress.ProgressIndicator
 import com.intellij.openapi.progress.ProgressManager
+import org.virtuslab.ideprobe.EndpointsConfig
 import scala.annotation.tailrec
 import scala.concurrent.duration._
 import org.virtuslab.ideprobe.Extensions._
+import org.virtuslab.ideprobe.handlers.IntelliJApi.getAwaitIdleConfig
```
lukaszwawrzyk, line 10: probably unused

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/IntelliJApi.scala
```diff
@@ -1,26 +1,31 @@
...
-import scala.concurrent.Future
+import scala.concurrent.{Await, Future}
 import scala.concurrent.duration.Duration
 import scala.reflect.ClassTag
+import pureconfig.generic.auto._
+import org.virtuslab.ideprobe.jsonrpc.PayloadJsonFormat._
 
-object IntelliJApi extends IntelliJApi
+object IntelliJApi extends IntelliJApi {
+  private var config = EndpointsConfig()
```
lukaszwawrzyk, line 18: This certainly is not part of intellij api utils, but a probe thing. Also object is here just to allow avoiding inheritance, nothing should be added to it ever.
Just create a new object in handlers like `ProbeConfig` with `initialize(config: String)` and `get(): Config` (we want general api for future and for extensions that might want to hook in to this config. BackgroundTasks should just `val awaitIdleConfig = ProbeConfig.get()[AwaitIdleConfig]("probe.endpoints.awaitIdle")`

api/src/main/scala/org/virtuslab/ideprobe/EndpointsConfig.scala
```diff
@@ -0,0 +1,15 @@
+package org.virtuslab.ideprobe
+
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
+
+case class EndpointsConfig(awaitIdle: EndpointsConfig.AwaitIdle)
+
+object EndpointsConfig {
+  def apply(): EndpointsConfig = EndpointsConfig(awaitIdle = AwaitIdle.Default)
+
+  case class AwaitIdle(initialWait: FiniteDuration, newTaskWait: FiniteDuration, checkFrequency: FiniteDuration)
```
lukaszwawrzyk, line 10: I'd put it outside of companion object and even put in a separate file as part of protocol. This is because it should also be an optional param to awaitIdle endpoint so that it can be configured per call from code.

lukaszwawrzyk (changes requested):

api/src/main/scala/org/virtuslab/ideprobe/protocol/AwaitIdle.scala
```diff
@@ -0,0 +1,9 @@
+package org.virtuslab.ideprobe.protocol
+
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
+
+case class AwaitIdle(initialWait: FiniteDuration, newTaskWait: FiniteDuration, checkFrequency: FiniteDuration)
```
lukaszwawrzyk, line 5: Maybe name it AwaitIdleParams, I think there is similar situation with building (check the class name there)

driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -53,7 +47,9 @@ final class InstalledIntelliJ(val root: Path, config: DriverConfig) {
       val connection = JsonRpcConnection.from(socket)
 
       val driver = ProbeDriver.start(connection)
-
+      val stringFromConfig = probeConfig.source.value
```
lukaszwawrzyk, line 50: it would be nice to put all the newly added logic into a method

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/BackgroundTasks.scala
```diff
@@ -18,16 +19,24 @@ object BackgroundTasks extends IntelliJApi {
     doWait()
   }
 
-  @tailrec
   private def doWait(): Unit = {
-    sleep(5.seconds)
+    val config = ProbeConfig.get()
```
lukaszwawrzyk, line 23: We'd like to have an overload of `awaitNone` that is public and takes AwaitIdleParams as parameter. Additionally current awaitIdle endpoint should be extended to accept Option[AwaitIdleParams] under the hood (driver should have 2 overloads (AwaitIdleParams, and 0 arg veriant).

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/ProbeConfig.scala
```diff
@@ -0,0 +1,13 @@
+package org.virtuslab.ideprobe.handlers
+
+import org.virtuslab.ideprobe.Config
+
+object ProbeConfig {
+  private var config = Config.Empty
+
+  def initialize(configContent: String): Unit = {
```
lukaszwawrzyk, line 8: It would be nicer if we had another overload that accepts Config

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/BackgroundTasks.scala
```diff
@@ -1,11 +1,12 @@
...
 import java.lang.reflect.Method
-import com.intellij.openapi.progress.ProgressIndicator
-import com.intellij.openapi.progress.ProgressManager
+import com.intellij.openapi.progress.{ProgressIndicator, ProgressManager}
+import org.virtuslab.ideprobe.Extensions._
+import org.virtuslab.ideprobe.protocol.AwaitIdle
 import scala.annotation.tailrec
 import scala.concurrent.duration._
-import org.virtuslab.ideprobe.Extensions._
+import pureconfig.generic.auto._
```
lukaszwawrzyk, line 9: make sure that the class that uses this import extends from ConfigFormat (I don't remember if IntelliJApi does)

api/src/main/scala/org/virtuslab/ideprobe/protocol/AwaitIdle.scala
MichalBednarz: There are BuildParams

driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
MichalBednarz: how about higher order function then

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/BackgroundTasks.scala
MichalBednarz: we have 3 awaitNone overloads now including one with option that seems a little bit dirty for me - check it out again please and tell me what you think now

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/BackgroundTasks.scala
MichalBednarz: that must be PayloadJsonFormat

PR #66, opened 2020-10-21
Change makeExecutable so that it gives the permission to everyone

MichalBednarz:


api/src/main/scala/org/virtuslab/ideprobe/ProbeExtensions.scala
```diff
@@ -111,7 +111,7 @@ trait ProbeExtensions {
     def makeExecutable(): Path = {
       import java.nio.file.attribute.PosixFilePermission._
       val attributes = Files.getPosixFilePermissions(path)
-      attributes.add(OWNER_EXECUTE)
+      attributes.add(OTHERS_EXECUTE)
```
lukaszwawrzyk, line 114: this is like changing `chmod u+x` to `chmod o+x` while we want `chmod +x` so we need 3 .add calls actually

lukaszwawrzyk (changes requested):

PR #70, opened 2020-10-26
Add configurable paths for probe related caches and working directories

MichalBednarz:


driver/sources/src/main/scala/org/virtuslab/ideprobe/WorkspaceProvider.scala
```diff
@@ -58,8 +58,11 @@ object WorkspaceProvider {
 
 sealed trait WorkspaceTemplate extends WorkspaceProvider {
 
-  override final def setup(): Path = {
-    val workspaceBase = Files.createTempDirectory("ideprobe-workspace")
+  override final def setup(paths: PathsConfig): Path = {
+    val dirName = "ideprobe-workspace"
+    val suffix = scala.util.Random.nextLong().abs.toString
+    val workspaceBase = paths.workspaces.resolve(dirName + suffix)
```
lukaszwawrzyk, line 64: It would be best to add to PathExtensions method like createTempDirectory(prefix: String = "", suffix: String = ""). I believe it is safer to use UUID.randomUUID() rather than Random.nextLong()

driver/sources/src/main/scala/org/virtuslab/ideprobe/config/PathsConfig.scala
```diff
@@ -0,0 +1,26 @@
+package org.virtuslab.ideprobe.config
+
+import java.nio.file.{Path, Paths}
+import org.virtuslab.ideprobe.config.PathsConfig.{cachePath, screenshotsPath, tmpDir}
+
+case class PathsConfig(instances: Path = tmpDir,
```
lukaszwawrzyk, line 6: Something is wrong in this file. Why instances and worksaces by default go directly to temp dir?
Btw as I shown in the issue in example code:
```
val targetCloneDir = fixture.paths.base.resolve("pants-repos").create().resolve(Hash.md5("$uri-$ref"))
```
The base path for probe paths should also be available in the code so that extensions can use it to create its files in the same place as core code.

Also note that we often have a separation of `Config` objects and domain objects in form of for example IntelliJFactory that has a method `from` (btw. probably we should rename all of these methods to `fromConfig`) that creates itself i.e. the domain object from config object. I think we should do a similar thing here. Have IdeProbePaths/TemporaryPaths or something and create it from PathsConfig. This is useful because of another thing I'd like to have:
someone can just say `probe.paths.base = "/mnt/gfs/ideprobe"` in their config, and all of the directories that depend on base path will go to a subdir of that one. If someone changes `probe.paths.cache = "/Users/uwawrzyk/.cache/ideprobe"`, the cache would go to my home dir, but other things will still go where the base dir says (tmp). This can be implemented easily if you make all of the in PathsConfig an Option[Path], and then the domain object would not have Options thanks to the mapping logic.

driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -95,7 +97,7 @@ final class InstalledIntelliJ(val root: Path, config: DriverConfig) {
         "IDEA_VM_OPTIONS" -> vmoptions.toString,
         "IDEA_PROPERTIES" -> ideaProperties.toString,
         "IDEPROBE_DRIVER_PORT" -> server.getLocalPort.toString,
-        "IDEPROBE_OUTPUT_DIR" -> Paths.get("/tmp/ideprobe/output").toString,
```
lukaszwawrzyk, line 98: Note that we have docker setup around this env var and hardcoded directory, so we should update it.

driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJFactory.scala
```diff
@@ -1,40 +1,35 @@
...
     installIntelliJ(version, root)
     installPlugins(allPlugins, root)
 
-    new InstalledIntelliJ(root, config)
+    new InstalledIntelliJ(root, paths, config)
+  }
+
+  private def createInstanceDirectory(version: IntelliJVersion): Path = {
+    val suffix = scala.util.Random.nextLong().abs.toString
+    val instanceDirectoryName = s"intellij-instance-${version.build}-$suffix"
```
lukaszwawrzyk, line 29: same here, reuse the createTempDirectory method.

lukaszwawrzyk (changes requested):

extensions/scala/driver/src/main/scala/org/virtuslab/ideprobe/scala/ScalaPluginExtension.scala
```diff
@@ -22,7 +22,7 @@ trait ScalaPluginExtension { this: IdeProbeFixture =>
         // scala plugin. This is why we delete one of them. We declare the scala-library as an
         // optional dependency with config file probePlugin/src/main/resources/META-INF/scala-plugin.xml
         // so that ideprobe plugin can be loaded regardless of the missing scala-library.
-        inteliJ.paths.plugins.resolve("ideprobe/lib/scala-library.jar").delete()
+        inteliJ.intellijPaths.plugins.resolve("ideprobe/lib/scala-library.jar").delete()
```
lukaszwawrzyk, line 25: As for naming, I believe RunningIntellijFixture (usually named as intelliJ in tests) should have the api:
intelliJ.paths : IntelliJPaths
intelliJ.probePaths / intelliJ.tempPaths : IdeProbePaths

project/CI.scala
```diff
@@ -32,6 +30,7 @@ object CI {
                       |
                       |export IDEPROBE_DISPLAY=xvfb
                       |export JAVA_HOME=/usr/lib/jvm/java-11-openjdk
+                      |export IDEPROBE_SCREENSHOTS_DIR=/tmp/screenshots
```
lukaszwawrzyk, line 33: I think, according also to my first comment here, this should be `/tmp/ideprobe/screenshots`

driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -95,7 +97,7 @@ final class InstalledIntelliJ(val root: Path, config: DriverConfig) {
...
+        "IDEPROBE_SCREENSHOTS_DIR" -> sys.env.getOrElse("IDEPROBE_SCREENSHOTS_DIR", paths.screenshots.toString),
```
lukaszwawrzyk, line 100: I didn't image this env var as some name hardcoded in probe. We already have configuration file where we can put it. Hocon supports env variables and optional config values. This place should just take `paths.screenshots` as is.
What we should change is not the test platform, but just the configuration of tests. For tests configured in HOCON, like scala extension tests and examples tests, it will be entry in configuration:
```
probe.paths.screenshots = ${?IDEPROBE_SCREENSHOTS_DIR}
```
this will overwrite existing screenshots configuration only if env var is set and leave default if it is not.
We configure probe primarily with hocon and it is quite convenient use.
ProbeDriverTests create probe programatically. We should add a method to fixture `withPaths` that will copy probe with specific paths. You will notice one problem when implementing it, that paths currently live in 2 objects: factory and fixture. I believe we should keep it in factory only, and add a getter `def` to fixture (or not even that, I am not sure if there is any need to use it directly from fixture....). Anyway the IdeProbePaths in IntelliJFactory needs to be replaced in `withPaths` method in the Fixture.

lukaszwawrzyk (changes requested):

driver/sources/src/main/scala/org/virtuslab/ideprobe/IdeProbePaths.scala
```diff
@@ -0,0 +1,30 @@
...
+object IdeProbePaths {
+  val TemporaryPaths: IdeProbePaths = {
+    from(PathsConfig())
+  }
+
+  def from(config: PathsConfig): IdeProbePaths = {
+    IdeProbePaths(config.base, config.instances, config.workspaces, config.screenshots, config.cache)
+  }
+
+  def apply(base: Option[Path],
```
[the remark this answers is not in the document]
MichalBednarz: inlined in what sense precisely?

driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
MichalBednarz: The problem related to the screenshots directory set up explicitly in the config file with the environment variable is that we will have to remember to provide it in for every config file used in our test suites.

driver/sources/src/main/scala/org/virtuslab/ideprobe/IntelliJFixture.scala
```diff
@@ -41,6 +37,11 @@ final case class IntelliJFixture(
     copy(config = newConfig)
   }
 
+  def withPaths(pathsConfig: PathsConfig): IntelliJFixture = {
```
lukaszwawrzyk, line 40: PathsConfig is a class that appears in config parsing part of probe, we should't leak it to the normal api. Just accept IdeProbePaths here

driver/sources/src/main/scala/org/virtuslab/ideprobe/IntelliJFixture.scala
```diff
@@ -66,11 +67,15 @@ final case class IntelliJFixture(
   def withWorkspace = new MultipleRunsIntelliJ(this)
 
   def setupWorkspace(): Path = {
-    val workspace = workspaceProvider.setup()
+    val workspace = workspaceProvider.setup(probePaths())
     afterWorkspaceSetup.foreach(_.apply(this, workspace))
     workspace
   }
 
+  def probePaths(): IdeProbePaths = {
```
lukaszwawrzyk, line 75: this is pure method, no need for ()

driver/sources/src/main/scala/org/virtuslab/ideprobe/config/PathsConfig.scala
```diff
@@ -0,0 +1,15 @@
...
+import java.nio.file.Path
+
+case class PathsConfig(base: Option[Path],
+                       instances: Option[Path],
+                       workspaces: Option[Path],
+                       screenshots: Option[Path],
+                       cache: Option[Path])
+
+object PathsConfig {
+  def apply(): PathsConfig = {
```
lukaszwawrzyk, line 12: Just put default parameters in the case class, or instead of `apply` create a Default val. But probably none of these is needed.

examples/src/test/resources/projects/librarymanagement.conf
```diff
@@ -10,3 +10,5 @@ modules {
 }
 
 runner = "ScalaTests"
+
+probe.paths.screenshots = ${?IDEPROBE_SCREENSHOTS_DIR}
```
lukaszwawrzyk, line 14: this should be in probe block in line 4-5, but actually is it needed at all if we include ideprobe.conf?

MichalBednarz: it's not necessary there

PR #73, opened 2020-10-30
Enable examples in CI

MichalBednarz:


odisseus:
Looks like the new test stage have timed out. Try increasing the time limit for that stage.

MichalBednarz:
> Looks like the new test stage has timed out. Try increasing the time limit for that stage.

It can also be the possible cause for the test suite failure, though `test-examples` in comparison with other test sets like `test-probe` behave in a strange way in a couple of ways, for example based on the CI logs: 

- plugins (bundled ideprobe plugin and external robot plugin) are attempted to be installed several times. This is at the same time the place where output stops to be produced. 

- IntelliJ is not being fetched (lack of `Fetching ....`) nor the appropriate logs like `Installing IntelliJVersion(...)` are present which is impossible based on the code that is being executed in the `IntelliJFactory`

The `examples` seem to work fine locally (with docker and without the docker).

MichalBednarz:
It appeared that the problem stems from behaviour of parameterized tests API (junit5 and jupiter-interface). Without any thorough analysis, it seems that the runner attempted to execute test cases concurrently. JUnit4 parameterized tests API works fine though.

.github/workflows/release.yml
```diff
@@ -0,0 +1,17 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
+      - 2020.3
```
lukaszwawrzyk, line 7: can be removed as 2020.3 is in master

PR #78, opened 2020-11-13
Endpoint to list the available test run configurations

odisseus:


driver/sources/src/main/scala/org/virtuslab/ideprobe/ProbeDriver.scala
```diff
@@ -222,6 +228,7 @@ class ProbeDriver(
     runTestsFromGenerated(runConfiguration, runnerToSelect = Some(runnerToSelect))
   }
 
+  // Is this method part of the API or should it be private???
   def runTestsFromGenerated(runConfiguration: TestScope, runnerToSelect: Option[String]): TestsRunResult = {
```
lukaszwawrzyk, line 232: It is part of the api. Useful for `runTestsFromGenerated(iJ.config[TestScope]("scope"), iJ.config.get[String]("runner"))`

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
```diff
@@ -182,6 +149,65 @@ object RunConfigurations extends IntelliJApi {
     new RunnerSettingsWithProcessOutput(settings)
   }
 
+  private def availableRunConfigurations(
+                                          scope: TestScope
+                                        )(implicit ec: ExecutionContext): java.util.List[ConfigurationFromContext] = {
+    BackgroundTasks.withAwaitNone {
```
MichalBednarz, line 155: Formatter behaves strangely in such cases, I would format it by hand for now.

driver/sources/src/main/scala/org/virtuslab/ideprobe/ProbeDriver.scala
MichalBednarz: Right, in case of `awaitIdle` we make the method private but it there this kind api appears to be useful if the runner name is not provided in config.

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
lukaszwawrzyk, line 155: I think it shouldn't be needed for this method.

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
```diff
@@ -182,6 +149,65 @@ object RunConfigurations extends IntelliJApi {
     new RunnerSettingsWithProcessOutput(settings)
   }
 
+  private def availableRunConfigurations(
+                                          scope: TestScope
+                                        )(implicit ec: ExecutionContext): java.util.List[ConfigurationFromContext] = {
```
lukaszwawrzyk, line 154: Maybe return a scala collection, so that on the use side it doesn't need to be converted

lukaszwawrzyk, line 154: is `ec` used in this method?

driver/sources/src/main/scala/org/virtuslab/ideprobe/ProbeDriver.scala
odisseus: Okay, but IMO these three methods are redundant. I think the API should have either a method with a non-optional second argument and another method with just one argument, or just one method with an optional second argument, or maybe just a two-argument method with some default token value for the second argument.

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
odisseus: Fixed

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
odisseus: Good point

odisseus: Removed

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
odisseus: Removed

MichalBednarz (approved):

lukaszwawrzyk (approved):

driver/sources/src/main/scala/org/virtuslab/ideprobe/ProbeDriver.scala
lukaszwawrzyk: We cover 2 use cases: generic, and simple calls. We should rather have an api that can be used conveniently by the users.

lukaszwawrzyk:
You can ignore current travis failure

PR #83, opened 2020-12-29
Encode UUIDs with Base 64

odisseus:
The motivation, purpose and implementation is exactly same as in https://github.com/scalacenter/bloop/pull/851.

Before: `/tmp/workspaces/ideprobe-workspacea5aabeda-930c-422e-8c68-ab9bce1bf81a/ws`
After: `/tmp/workspaces/ideprobe-workspace-uAVmccQ9SZ6SVESwNdmDSw==/ws`

### DIFF
--- a/api/src/main/scala/org/virtuslab/ideprobe/ProbeExtensions.scala
+++ b/api/src/main/scala/org/virtuslab/ideprobe/ProbeExtensions.scala
@@ -13,7 +13,6 @@ import java.nio.file.StandardCopyOption
 import java.nio.file.StandardOpenOption
 import java.nio.file.SimpleFileVisitor
 import java.nio.file.attribute.BasicFileAttributes
-import java.util.UUID
 import java.util.zip.ZipInputStream
 import scala.util.Failure
 import scala.util.Try
@@ -61,8 +60,7 @@ trait ProbeExtensions {
     }
 
     def createTempDirectory(prefix: String): Path = {
-      val suffix = UUID.randomUUID.toString
-      path.resolve(prefix + suffix)
+      path.resolve(s"$prefix-${UUIDs.randomUUID()}")
     }
 
     def copyTo(target: Path): Path = {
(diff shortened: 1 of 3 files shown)


api/src/main/scala/org/virtuslab/ideprobe/util/UUIDUtil.scala
```diff
@@ -0,0 +1,19 @@
+package org.virtuslab.ideprobe.util
+
+import java.nio.ByteBuffer
+import java.util.{Base64, UUID}
+
+object UUIDUtil {
```
lukaszwawrzyk, line 6: Can we adjust names & packages, so that it can be referred as:
`org.virtuslab.ideprobe.UUIDs.randomUUID()`

api/src/main/scala/org/virtuslab/ideprobe/util/UUIDUtil.scala
```diff
@@ -0,0 +1,19 @@
...
+  // Copied from https://stackoverflow.com/a/29836273
+
+  def toBase64String(uuid: UUID): String = {
+    val bb = ByteBuffer.wrap(new Array[Byte](16))
+    bb.putLong(uuid.getMostSignificantBits)
+    bb.putLong(uuid.getLeastSignificantBits)
+    Base64.getUrlEncoder.encodeToString(bb.array)
+  }
+
+  def randomUUID: String = toBase64String(UUID.randomUUID)
```
lukaszwawrzyk, line 17: can you add empty `()` here and at call site? So signify that it is impure

lukaszwawrzyk (approved):

api/src/main/scala/org/virtuslab/ideprobe/util/UUIDUtil.scala
odisseus: Done

api/src/main/scala/org/virtuslab/ideprobe/util/UUIDUtil.scala
odisseus: Done

PR #84, opened 2020-12-29
Set up CI tests using Github Actions

odisseus:


lukaszwawrzyk:
It would be very helpful if this worked. For the time being we could use gh actions to run tests and travis to release, eventually switching fully to gh actions.

lukaszwawrzyk (approved):
Let's remove testing from travis and just leave publishing there

odisseus:
What about prerequisites for releasing? How would Travis know that the tests have passed?

tpasternak (approved):

lukaszwawrzyk:
> What about prerequisites for releasing? How would Travis know that the tests have passed?

There are problems with tests stability in this repo and they most frequently were failing when they shouldn't and we wanted a release anyway so we would run them multiple times until they pass. Currently many of tests are disabled. The tests should be fixed to be stable and enabled back. But regardless of that, we want to be able to merge a PR and use a version of probe from such PR in other repo, for this reason it is ok to publish snapshot if the repo compiles (i.e. run publish always on merge to master).

PR #85, opened 2021-01-11
Build artifacts with ide-probe

tpasternak:


probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/Projects.scala
```diff
@@ -130,4 +131,17 @@ object Projects extends IntelliJApi {
   }
 
   private def toRef(project: Project): ProjectRef = ProjectRef(project.getName)
+
+  def buildArtifact(projectRef: ProjectRef, artifactName: String): Unit = {
+    val defaultProject: Project = projectRef match {
```
lukaszwawrzyk, line 136: equivalent of Projects.resolve

probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/Projects.scala
```diff
@@ -130,4 +131,17 @@ object Projects extends IntelliJApi {
   }
 
   private def toRef(project: Project): ProjectRef = ProjectRef(project.getName)
+
+  def buildArtifact(projectRef: ProjectRef, artifactName: String): Unit = {
```
lukaszwawrzyk, line 135: I'd move this to already existing object called "Builds" in same package.

lukaszwawrzyk (approved):

lukaszwawrzyk (approved):

PR #103, opened 2021-03-03
Wrapping fixtures variables with Options

sellophane:
Wrapping said vars with Options prevents NullPointerExceptions on `teardown()` in situation when variables are not correctly initialised during `setup()`

### DIFF
--- a/driver/sources/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
+++ b/driver/sources/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
@@ -8,12 +8,14 @@ import org.virtuslab.ideprobe.reporting.AfterTestChecks
 
 trait RunningIntelliJPerSuiteBase {
 
-  private var workspace: Path = _
-  private var installed: InstalledIntelliJ = _
-  private var running: RunningIde = _
-  private var runningIntelliJFixture: RunningIntelliJFixture = _
+  private var workspace: Option[Path] = None
+  private var installed: Option[InstalledIntelliJ] = None
+  private var running: Option[RunningIde] = None
+  private var runningIntelliJFixture: Option[RunningIntelliJFixture] = None
 
-  final def intelliJ: RunningIntelliJFixture = runningIntelliJFixture
+  final def intelliJ: RunningIntelliJFixture = runningIntelliJFixture.getOrElse(
+    error("Intellij Fixture not initialized")
+  )
 
   protected def baseFixture: IntelliJFixture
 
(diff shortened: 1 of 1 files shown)


driver/sources/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
```diff
@@ -8,12 +8,15 @@ import org.virtuslab.ideprobe.reporting.AfterTestChecks
...
-  private var running: RunningIde = _
-  private var runningIntelliJFixture: RunningIntelliJFixture = _
-
-  final def intelliJ: RunningIntelliJFixture = runningIntelliJFixture
+  private var workspace: Option[Path] = None
+  private var installed: Option[InstalledIntelliJ] = None
+  private var running: Option[RunningIde] = None
+  private var runningIntelliJFixture: Option[RunningIntelliJFixture] = None
+
+  final def intelliJ: RunningIntelliJFixture = runningIntelliJFixture match {
```
lukaszwawrzyk, line 16: `runningIntelliJFixture.getOrElse(error("IntelliJ fixture not initialized"))`. 
`getOrElse` for more idiomatic handling options. 
As for `java.lang.Error` it signifies some more serious causes, like OutOfMemoryError, LinkageError, NoSuchMethodError, AssertionError etc. We just want a RuntimeException. Inside ide-probe code base, I think everywhere we use the `org.virtuslab.ideprobe.error` function, that just does `throw new RunimeException(msg)`, but is shorter to write, and whenever we need to introduce some special type of exception, we will be able to change it in one place.

driver/sources/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
```diff
@@ -22,33 +25,43 @@ trait RunningIntelliJPerSuiteBase {
   protected def afterAll(): Unit = ()
 
   def setup(): Unit = {
-    workspace = baseFixture.setupWorkspace()
-    installed = baseFixture.installIntelliJ()
-    running = baseFixture.startIntelliJ(workspace, installed)
-    runningIntelliJFixture = new RunningIntelliJFixture(workspace, running.probe, baseFixture.config, installed.paths)
+    workspace = Some(baseFixture.setupWorkspace())
```
lukaszwawrzyk, line 28: Maybe also imperfect, but I would structure this method differently. Options and handling options should not be visible throughout this method. Current implementation below this comment suggest that `workspace`, `installed` or anything can be null/None and it needs to be accessed through for comprehension. But actually it doesn't, none of the setup methods return option, and they don't return null (and they really, really shouldn't). I would separate **initialization code** (the original code) from assigning fields to what they expect.

```scala
    val workspace = baseFixture.setupWorkspace()
    this.workspace = Some(workspace)
    val installed = baseFixture.installIntelliJ()
    this.installed = Some(installed)
    val running = baseFixture.startIntelliJ(workspace, installed)
    this.running = Some(running)
    this.runningIntelliJFixture = Some(new RunningIntelliJFixture(workspace, running.probe, baseFixture.config, installed.paths))
    beforeAll()
```

It is not that nice because of assignment and logic are mixed with each other, but it is necessary in case setup fails at any step, so that it will be cleaned up. But hopefully it is more clear.

Another alternative would be to leave fields as they were and leave this method as it was, just in the teardown do the nullchecks in form of `Option(running).foreach(baseFixture.closeIntellij)`. I think either options is fine, all are imperfect, but I think more readable than current.

lukaszwawrzyk (changes requested):

lukaszwawrzyk (changes requested):
The comment at line 16 is not resolved, even though marked as resolved ;p

sellophane:
> The comment at line 16 is not resolved, even though marked as resolved ;p

My bad, I resolved it partially. Now should be okay.

lukaszwawrzyk (approved):

PR #105, opened 2021-03-11
Adding semi-automatic configuration parser for TestScope

sellophane:
This PR adds semi-automatic PureConfig readers for TestScope and all it's components

### DIFF
--- a/core/api/src/main/scala/org/virtuslab/ideprobe/protocol/ModuleRef.scala
+++ b/core/api/src/main/scala/org/virtuslab/ideprobe/protocol/ModuleRef.scala
@@ -1,9 +1,15 @@
 package org.virtuslab.ideprobe.protocol
 
+import org.virtuslab.ideprobe.ConfigFormat
+import pureconfig.ConfigConvert
+import pureconfig.generic.semiauto.deriveConvert
+
 final case class ModuleRef(name: String, project: ProjectRef = ProjectRef.Default)
 
-object ModuleRef {
+object ModuleRef extends ConfigFormat {
   def apply(name: String, project: String): ModuleRef = {
     ModuleRef(name, ProjectRef(project))
   }
+
+  implicit val moduleRefConvert: ConfigConvert[ModuleRef] = deriveConvert[ModuleRef]
 }
--- a/core/api/src/main/scala/org/virtuslab/ideprobe/protocol/ProjectRef.scala
+++ b/core/api/src/main/scala/org/virtuslab/ideprobe/protocol/ProjectRef.scala
@@ -1,10 +1,34 @@
 package org.virtuslab.ideprobe.protocol
 
+import com.typesafe.config.ConfigValueFactory
+import org.virtuslab.ideprobe.ConfigFormat
+import pureconfig.generic.semiauto.{deriveReader, deriveWriter}
+import pureconfig.{ConfigReader, ConfigWriter}
+
 sealed trait ProjectRef
 
-object ProjectRef {
+object ProjectRef extends ConfigFormat {
   case object Default extends ProjectRef
   case class ByName(name: String) extends ProjectRef
 
   def apply(name: String): ProjectRef = ByName(name)
+
+  private val defaultProjectRefConfigReader: ConfigReader[Default.type] =
+    ConfigReader.fromStringOpt(s => if (s.toLowerCase == "default") Some(Default) else None)
+
+  implicit val projectRefConfigReader: ConfigReader[ProjectRef] = {
+    possiblyAmbiguousAdtReader(
+      deriveReader[ByName],
+      defaultProjectRefConfigReader
+    )
+  }
+
+  implicit val projectRefConfigWriter: ConfigWriter[ProjectRef] = {
+    possiblyAmbiguousAdtWriter(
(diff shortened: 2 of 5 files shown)


api/src/test/scala/org/virtuslab/ideprobe/TestScopeParsingTest.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import org.virtuslab.ideprobe.protocol.TestScope
+import pureconfig.ConfigSource
+
+@RunWith(classOf[JUnit4])
+final class TestScopeParsingTest {
+
+  private def testParsing(confString: String): TestScope = {
+    val conf = ConfigFactory.parseString(confString)
```
lukaszwawrzyk, line 14: We have wrapped this api in `Config` class. It would be better to use the wrapper, in case we ever replaced the underlying implemenation.
This would be `Config.fromString(...).as[TestScope]`

api/src/test/scala/org/virtuslab/ideprobe/TestScopeParsingTest.scala
```diff
@@ -0,0 +1,49 @@
...
+  @Test
+  def parseModule(): Unit = {
+    val scope = testParsing("""{ module: { name: "moduleName" }}""")
+    assert(scope.isInstanceOf[TestScope.Module])
+  }
+
+  @Test
+  def parseDirectory(): Unit = {
+    val scope = testParsing("""{ module: { name: "moduleName" }, directory-name: "/tmp" }""")
+    assert(scope.isInstanceOf[TestScope.Directory])
```
lukaszwawrzyk, line 29: Lets make sure values are also correct, it will additionally make it more eye pleasing, without `instanceOf`.
assertEquals(scope, TestScope.Directory(directoryName = "/tmp", ModuleRef("moduleName"))

lukaszwawrzyk (changes requested):

core/api/src/main/scala/org/virtuslab/ideprobe/protocol/ModuleRef.scala
```diff
@@ -1,9 +1,17 @@
...
+
 final case class ModuleRef(name: String, project: ProjectRef = ProjectRef.Default)
 
-object ModuleRef {
+object ModuleRef extends ConfigFormat {
   def apply(name: String, project: String): ModuleRef = {
     ModuleRef(name, ProjectRef(project))
   }
+
+  implicit val moduleRefReader: ConfigReader[ModuleRef] = deriveReader[ModuleRef]
```
lukaszwawrzyk, line 14: I think here is deriveConvert that derives ConfigConvert which does both reading and writing

core/api/src/main/scala/org/virtuslab/ideprobe/protocol/ProjectRef.scala
```diff
@@ -1,10 +1,42 @@
...
-object ProjectRef {
+object ProjectRef extends ConfigFormat {
   case object Default extends ProjectRef
   case class ByName(name: String) extends ProjectRef
 
   def apply(name: String): ProjectRef = ByName(name)
+
+  private val defaultProjectRefConfigReader: ConfigReader[ProjectRef.Default.type] = (cursor: ConfigCursor) => {
+    cursor.asString match {
+      case Left(failures) => Left[ConfigReaderFailures, ProjectRef.Default.type](failures)
```
lukaszwawrzyk, line 19: Step 1. You don't need to provide types for Left and Right, by the type of val in line 17 it is known what kind of Either is expected as return type. Compiler will just check if it seems ok. 
Step 2. left is noop, you don't change anything, we just want to transform Right, so you can just use `map` instead of pattern matching. Actually `flatMap`, as you might return Left or Right.
Step 3. use built in factory function:
```
ConfigReader.fromStringOpt(s => if (s.toLowerCase == "default") Some(Default) else None)
```

core/api/src/main/scala/org/virtuslab/ideprobe/protocol/ProjectRef.scala
```diff
@@ -1,10 +1,42 @@
...
+  }
+
+  implicit val projectRefConfigReader: ConfigReader[ProjectRef] = {
+    possiblyAmbiguousAdtReader(
+      deriveReader[ByName],
+      defaultProjectRefConfigReader
+    )
+  }
+
+  implicit val projectRefConfigWriter: ConfigWriter[ProjectRef] = ConfigWriter.fromFunction {
```
lukaszwawrzyk, line 38: Writer for ByName should be derived. ConfigValueFactory only accepts basic types, it fails for your tuple.
```
  private val defaultProjectRefConfigReader: ConfigReader[ProjectRef.Default.type] =
    ConfigReader.fromStringOpt(s => if (s.toLowerCase == "default") Some(Default) else None)

  implicit val projectRefConfigReader: ConfigReader[ProjectRef] = {
    possiblyAmbiguousAdtReader(
      deriveReader[ByName],
      defaultProjectRefConfigReader
    )
  }
```

PR #106, opened 2021-03-15
Implement WaitLogic

lukaszwawrzyk:


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/wait/ConstantWaiting.scala
```diff
@@ -0,0 +1,27 @@
...
+    duration: FiniteDuration,
+    waitCondition: ProbeDriver => WaitDecision
+) extends WaitLogic {
+
+  def this(duration: FiniteDuration) = {
+    this(duration, _ => WaitDecision.Done)
+  }
+
+  override def await(driver: ProbeDriver): Unit = {
+    waitCondition(driver)
```
tpasternak, line 16: We name this param `waitCondition` but we don't check its result

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/wait/WaitingWithEnsurePeriod.scala
```diff
@@ -0,0 +1,71 @@
+package org.virtuslab.ideprobe
+package wait
+
+import scala.annotation.tailrec
+import scala.concurrent.duration.{Duration, FiniteDuration}
+
+class WaitingWithEnsurePeriod(
```
tpasternak, line 7: This one is very complex, I think it needs some documentation. Apart from this, IMO we could rename it to something like: `WaitingUntilConditionIsStable`

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/wait/ExecutionBackoff.scala
```diff
@@ -0,0 +1,23 @@
+package org.virtuslab.ideprobe.wait
+
+import scala.concurrent.duration.FiniteDuration
+
+class ExecutionBackoff(maxFrequency: FiniteDuration) {
```
tpasternak, line 5: Personally, I would rename it to something like `Throttle`

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/wait/WaitLimit.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  def check(): Unit = {
+    check(s"Waiting limit of $limit exceeded.")
+  }
+
+  def check(errorMessage: String): Unit = {
+    val now = System.currentTimeMillis()
+    val duration = (now - startTime).millis
+    if (duration >= limit) {
+      error(errorMessage)
```
tpasternak, line 21: Where does `error` method come from?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/WaitLogic.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  def and(waitCondition: ProbeDriver => WaitDecision): WaitLogic
+
+  def or(waitCondition: ProbeDriver => WaitDecision): WaitLogic
+
+  def doWhileWaiting(maxFrequency: FiniteDuration)(code: => Unit): WaitLogic = {
+    val backoff = new ExecutionBackoff(maxFrequency)
+    or { _ =>
+      backoff.execute { code }
+      WaitDecision.Done
```
tpasternak, line 18: I'm not sure if I understand it correctly. Imo we should either return `WaitDecision.Keep`, or call it with `and`. Otherwise it will always terminate until one check (because we're doing something like `or true`)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/wait/ConstantWaiting.scala
lukaszwawrzyk: This is constant waiting. I execute condition for consistency and potential side effects. It waits for constant time, no matter what.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/wait/WaitLimit.scala
lukaszwawrzyk: org.virtuslab.ideprobe.error

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/WaitLogic.scala
lukaszwawrzyk: WaitDecision answers the question `needToWait`. Perhaps I should invert this to be more intuitive.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/WaitDecision.scala
```diff
@@ -0,0 +1,42 @@
+package org.virtuslab.ideprobe
+
+sealed trait WaitDecision {
+  def needToWait: Boolean
+  def message: Option[String]
+  def ||(other: WaitDecision): WaitDecision = WaitDecision.or(this, other)
+  def &&(other: WaitDecision): WaitDecision = WaitDecision.and(this, other)
```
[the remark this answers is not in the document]
lukaszwawrzyk: It can be handled with not-that-bad
```
a || b || c
List(x, y, z).reduce(_ || _)
```
if needed, but one would rather want to `or` on the `WaitLogic` than on `WaitDecision`, but `or` and `and` on that class is not `(WaitLogic, WaitLogic) => WaitLogic` function, but `(WaitLogic, ProbeDiver => WaitDecision) => WaitLogic`

lukaszwawrzyk:
@odisseus  I resolved conversation accidentally. How is `ChainedWaiting` different than `WaitDecision.or`. There is a couple of things here:
`WaitDecision` - more or less, a domain specific boolean that informs if one should wait or not
`WaitCondition` - not in code, but conceptually is there, it is: `ProbeDriver => WaitDecision`. It actually decides what are we waiting for, and this is actually what we can `or` or `and`. For example we are waiting for bg task to appear or error appear in build panel.
`WaitLogic` - describes how we are waiting: 
 - `ConstantWaiting`: always for given constant
 - `BasicWaiting`: while condition says so, checking every specified duration
 - `WaitingWIthEnsurePeriod`: like basic, but also checking after waiting would end to make sure it is stable
 - others...
`WaitLogic` contains the `WaitCondition` and executes it in specific way described above and allows to modify the WaitCondition inside through `or` and `and` method. 

Pseudocode:
```
val waitForStart = BasicWaiting(driver => backgroundTaskAppears("Indexing"))
val waitForEnd =  WaitingWIthEnsurePeriod(driver => backgroundTaskDisappears("Indexing"))

// this doesn't make sense, how would we wait for both such conditions in the same time and using two different waiting strategies, this code is not possible to happen with current implementation as you cannot `or` two `WaitLogic` objects.
waitForStart or waitForEnd

// This one makes sense, we wait for first condition in some way, andThen wait for other condition in a different way.
// `andThen` method is not there, but it can be expressed with ChainedWaiting as WaitLogic.chained(waitForStart, waitForEnd)
waitForStart andThen waitForEnd
```

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/wait/WaitLimit.scala
```diff
@@ -0,0 +1,25 @@
+package org.virtuslab.ideprobe
+package wait
```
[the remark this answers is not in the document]
lukaszwawrzyk: Separating packages this way allows to skip imports from `org.virtuslab.ideprobe`

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/wait/WaitingWithEnsurePeriod.scala
lukaszwawrzyk: done

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/wait/ExecutionBackoff.scala
lukaszwawrzyk: done

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/WaitLogic.scala
lukaszwawrzyk: done

tpasternak (approved):

PR #108, opened 2021-03-18
Modifying outputs for docker runs and base path

sellophane:


odisseus:
This should probably be a draft.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IdeProbePaths.scala
```diff
@@ -11,7 +11,7 @@ object IdeProbePaths {
   }
 
   def from(config: PathsConfig): IdeProbePaths = {
-    val basePath = config.base.getOrElse(Paths.get(System.getProperty("java.io.tmpdir")))
+    val basePath = config.base.getOrElse(Paths.get(System.getProperty("java.io.tmpdir") + "/ide_probe"))
```
lukaszwawrzyk, line 14: `Paths.get(System.getProperty("java.io.tmpdir")).resolve("ide-probe")` would be more clean. Also I meant `-` as in repo name

ci/tests/run
```diff
@@ -1,8 +1,8 @@
 #!/bin/sh
 
 DOCKER_IMAGE=testimage:local
-DOCKER_DIRECTORY=/tmp/ideprobe/screenshots
-HOST_DIRECTORY=/tmp/ideprobe/screenshots
+DOCKER_DIRECTORY=/tmp/ide_probe
+HOST_DIRECTORY=/tmp/ide_probe
```
lukaszwawrzyk, line 5: also fix paths here

lukaszwawrzyk:
It could be a separate PR but we discussed adding screenshot upload to ide-probe similar to pants tests.

Something like below, but the name would need to include tests-set and scala-version to be unique and path needs to be corrected.
```
      - name: Upload screenshots
        if: always()
        uses: actions/upload-artifact@v2
        with:
          name: screenshots-${{ matrix.suite-name }}
          path: /tmp/ideprobe/output
```

PR #110, opened 2021-03-19
Bazel build

lukaszwawrzyk:


extensions/bazel/driver/src/main/scala/org/virtuslab/ideprobe/bazel/BazelProbeDriver.scala
```diff
@@ -1,18 +1,138 @@
...
+
+    // set project view configuration file
+    robot.clickRadioButton("Import project view file")
+    robot.find(query.div("name" -> "projectview-file-path-field")).setText(projectView.toString)
+    robot.clickButton("Next")
+
+    // complete import wizard
+    robot.clickButton("Finish")
+
+    driver.await(waitLogic)
```
tpasternak, line 98: Should this be a parameter? Should we leave decision about wait logic to the user? Personally, if I was a user,  I would like the implementation to make the right choice

tpasternak (approved):

extensions/bazel/driver/src/main/scala/org/virtuslab/ideprobe/bazel/BazelProbeDriver.scala
lukaszwawrzyk: When we open project we need some sort of waiting, this is for sure. Default wait logic will wait for indexing and python skeletons update to complete. In benchmarks for example we just wait for the sync to complete by looking at build panel. Also specific parameters of waiting, like basicCheckFrequency in default version are quite long (5 seconds), which seemed to help us with flakiness, but in git-machete they care more about fast test execution and prefer rather tighter check loop and as they import a small project it doesn't matter that much.

PR #113, opened 2021-03-26
Adding plugin version switcher to probe drivers

sellophane:
Adding both scala 2.12 and 2.13 plugins to drivers archives and enabling runtime pick based on IntelliJ version

### DIFF
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/InternalPlugins.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/InternalPlugins.scala
@@ -1,11 +1,15 @@
 package org.virtuslab.ideprobe.dependencies
 
-import org.virtuslab.ideprobe.BuildInfo
+import org.virtuslab.ideprobe.{BuildInfo, IntelliJFixture}
 
 object InternalPlugins {
-  val probePlugin: Plugin = bundle("ideprobe")
 
-  val all = Seq(probePlugin)
+  def bundleCross(name: String, intelliJVersion: IntelliJVersion): Plugin =
+    Plugin.BundledCrossVersion(name, intelliJVersion.compatibleScalaVersion, BuildInfo.version)
 
-  def bundle(name: String): Plugin = Plugin.Bundled(s"$name-${BuildInfo.version}.zip")
+  def probePluginForIntelliJ(intelliJVersion: IntelliJVersion): Plugin = bundleCross("ideprobe", intelliJVersion)
+
+  def installCrossVersionPlugin(name: String): IntelliJFixture => IntelliJFixture = { fixture =>
+    fixture.withPlugin(bundleCross(name, fixture.version))
+  }
 }
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/PluginResolver.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/PluginResolver.scala
@@ -1,7 +1,7 @@
 package org.virtuslab.ideprobe.dependencies
 
 import org.virtuslab.ideprobe.config.DependenciesConfig
-import org.virtuslab.ideprobe.dependencies.Plugin.Bundled
+import org.virtuslab.ideprobe.dependencies.Plugin.{Bundled, BundledCrossVersion}
 
 object PluginResolver {
   val Official = PluginResolver("https://plugins.jetbrains.com/plugin/download")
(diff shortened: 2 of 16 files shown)


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/PluginResolver.scala
```diff
@@ -29,13 +39,13 @@ object PluginResolver {
         case Plugin.Versioned(id, version, None) =>
           Dependency(s"$uri?pluginId=$id&version=$version")
 
+        case BundledCrossVersionZip(name, scalaVersion, version) => {
+          val suffix = if (version.nonEmpty) s"-$version" else ""
```
lukaszwawrzyk, line 43: Why would version be empty?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/InternalPlugins.scala
```diff
@@ -3,9 +3,11 @@ package org.virtuslab.ideprobe.dependencies
 import org.virtuslab.ideprobe.BuildInfo
 
 object InternalPlugins {
-  val probePlugin: Plugin = bundle("ideprobe")
+  def probePlugin(version: String): Plugin = bundleCross("ideprobe", version)
 
-  val all = Seq(probePlugin)
+  def allForVersion(version: String) = Seq(probePlugin(version))
 
   def bundle(name: String): Plugin = Plugin.Bundled(s"$name-${BuildInfo.version}.zip")
```
lukaszwawrzyk, line 10: It would be a good method to use before, but I forgot to use it, now it is rather useless so you could delete it.
Also bundleCross has little use, so you might just want to inline that.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IntelliJFixture.scala
```diff
@@ -57,6 +57,11 @@ final case class IntelliJFixture(
     copy(plugins = plugin +: plugins)
   }
 
+  def withBundledCrossVersion(name: String, pluginVersion: String): IntelliJFixture = {
+    val plugin = Plugin.BundledCrossVersionZip(name, version.compatibleScalaVersion, pluginVersion)
+    copy(plugins = plugin +: plugins)
```
lukaszwawrzyk, line 62: This line can be implemented as `withPlugin(plugin)`

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IntelliJFixture.scala
```diff
@@ -57,6 +57,11 @@ final case class IntelliJFixture(
     copy(plugins = plugin +: plugins)
   }
 
+  def withBundledCrossVersion(name: String, pluginVersion: String): IntelliJFixture = {
```
lukaszwawrzyk, line 60: This is most user facing api, perhaps we could move this setup somewhere else, to still be useful to us, but in more adequate place.
I thought maybe of something like putting it in InternalPlugins (the place I asked to cleanup from helper method with little use now).
```scala
object InternalPlugins {
  def installCrossVersionPlugin(name: String): IntelliJFixture => IntelliJFixture = {
    fixture => fixture.withPlugin(Plugin.BundledCrossVersion(name, fixture.version.compatibleScalaVersion, BuildInfo.version))
  }
}
```
It would be a useful utility function for us, as it would handle all versions at once. We don't need to care about being able to change version to something else, as all internal plugins will have the same version. 

In extension traits it can be used as:
```scala
  registerFixtureTransformer(InternalPlugins.installCrossVersionPlugin("ideprobe-pants"))
```

lukaszwawrzyk (changes requested):

lukaszwawrzyk:
As for tests, I was a little confused at standup, but to explain it again:
We run github actions for 2 suites: core (probe) and scala. Both run for 2.12 and 2.13. But somewhere in sbt/ci related code scala 2.12 run is hacked to not run any tests, this is because we run tests for 2020.3 and it doesn't work with 2.12 scala.
If you find this hacky place and get rid of this code, it will just start running scala tests on 2.12 also. If all tests pass, it will prove that we can run 2020.3 tests both on scala 2.12 and 2.13. 
Perhaps it would be nice to prove that we can run 2020.2 tests on 2.13 but it is not as important, especially, that it is not a scenario we need to support.
To summarize, it should be enough to enable all tests and make sure they pass.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJFactory.scala
```diff
@@ -19,7 +20,7 @@ final class IntelliJFactory(
   def create(version: IntelliJVersion, plugins: Seq[Plugin]): InstalledIntelliJ = {
     val root = createInstanceDirectory(version)
 
-    val allPlugins = InternalPlugins.all ++ plugins
+    val allPlugins = probePluginForIntelliJ(version) +: plugins
```
lukaszwawrzyk, line 23: If we would keep the `InternalPlugins` it would be more clear that function comes from outside rather than the base trait (specifically it is more important for the PluginExtension traits.

lukaszwawrzyk:
Once tests are fixed it is good to merge

lukaszwawrzyk (approved):

PR #130, opened 2021-05-13
Enabled usage of existing IntelliJ instances

kszydlo-vl:


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJFactory.scala
```diff
@@ -103,4 +107,16 @@ object IntelliJFactory {
     val dependencyProvider = new DependencyProvider(intelliJDependencyProvider, pluginDependencyProvider)
     new IntelliJFactory(dependencyProvider, paths, driverConfig)
   }
+
+  def version(root: Path): IntelliJVersion = {
```
MichalBednarz, line 111: logic duplicated in IntelliJProvider

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
```diff
@@ -0,0 +1,93 @@
...
+  override def setup(): InstalledIntelliJ = {
+    val pluginsDir = path.resolve("plugins")
+    val backupDir = Files.createTempDirectory(path, "plugins")
+    copyFolder(pluginsDir, backupDir)
+    factory.installPlugins(plugins, path)
+
+    new LocalIntelliJ(path, factory.paths, factory.config, backupDir)
+  }
+
+  private def copyFolder(source: Path, target: Path): Unit = {
```
MichalBednarz, line 34: put it in ProbeExtensions?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
```diff
@@ -0,0 +1,93 @@
...
+import org.virtuslab.ideprobe.dependencies.{IntelliJVersion, Plugin}
+
+import java.nio.file.attribute.BasicFileAttributes
+import scala.io.Source
+
+trait IntelliJProvider {
+  def factory: IntelliJFactory
+  def setup(): InstalledIntelliJ
+  def version: IntelliJVersion
+  def withFactory(factory: IntelliJFactory): IntelliJProvider
```
MichalBednarz, line 17: method implementation is same in both child classes, can be implemented in here?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
```diff
@@ -0,0 +1,96 @@
+package org.virtuslab.ideprobe.config
```
MichalBednarz, line 1: why not in ide.intellij?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
kszydlo-vl: This is tricky, as `withFactory` relies on `copy` method in case classes, therefore it's impossible to provide a completely subtype-agnostic implementation here. One possible idea would be something along this line:
```
final def withFactory(factory: IntelliJFactory): IntelliJProvider = this match {
  case any: ExistingIntelliJ => any.copy(factory = factory)
  case any: DefaultIntelliJ => any.copy(factory = factory)
}
```
however this solution's elegance is disputable.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
lukaszwawrzyk: Such method probably already exists in PathExtension

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
```diff
@@ -0,0 +1,96 @@
...
+          Files.copy(file, target.resolve(source.relativize(file)))
+          FileVisitResult.CONTINUE
+        }
+      }
+    )
+  }
+
+  override lazy val version: IntelliJVersion = {
+    case class ProductInfo(version: String, buildNumber: String)
+    val productInfo = path.resolve("product-info.json").toFile
```
lukaszwawrzyk, line 54: path.resolve(...).content()

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
```diff
@@ -0,0 +1,96 @@
...
+case class ExistingIntelliJ(
+    path: Path,
+    plugins: Seq[Plugin],
+    factory: IntelliJFactory
+) extends IntelliJProvider {
+  override def setup(): InstalledIntelliJ = {
+    val pluginsDir = path.resolve("plugins")
+    val backupDir = Files.createTempDirectory(path, "plugins")
+    copyFolder(pluginsDir, backupDir)
+    factory.installPlugins(plugins, path)
```
MichalBednarz, line 29: are all the required plugins installed? Take a look at `factory.create`. It takes plugins from `InternalPlugins.probePluginForIntelliJ(version)`. How about plugins from `IntelliJFixture`?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -3,16 +3,18 @@ package org.virtuslab.ideprobe.ide.intellij
...
 import org.virtuslab.ideprobe.config.DriverConfig
 import org.virtuslab.ideprobe.jsonrpc.JsonRpcConnection
+
 import scala.concurrent.{ExecutionContext, blocking}
 
-final class InstalledIntelliJ(val root: Path, probePaths: IdeProbePaths, config: DriverConfig) {
+sealed abstract class InstalledIntelliJ(root: Path, probePaths: IdeProbePaths, config: DriverConfig) {
+  def cleanup(): Unit
+
   val paths: IntelliJPaths = new IntelliJPaths(root, config.headless)
```
MichalBednarz, line 18: Existing IntelliJ instance can have specific directory config (specified in `*.properties`). We want those paths to be manually configurable. (or maybe we can automatically detect the existing directory structure, for example by reading those locations from existing `*.properties` or by providing default structure?)

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJFactoryTest.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.runners.JUnit4
+import org.junit.Test
+import org.virtuslab.ideprobe.ide.intellij.IntelliJFactory
+
+import scala.io.Source
+
+@RunWith(classOf[JUnit4])
+final class IntelliJFactoryTest {
+
+
```
MichalBednarz, line 17: nitpick: remove unnecessary empty lines

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IntelliJFixture.scala
```diff
@@ -58,35 +64,39 @@ final case class IntelliJFixture(
   }
```
lukaszwawrzyk, line 64: The method withPlugin is broken now as `plugins` is unused

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -128,3 +130,24 @@ final class InstalledIntelliJ(val root: Path, probePaths: IdeProbePaths, config:
     builder.start()
   }
 }
+
+final class LocalIntelliJ(
+    val root: Path,
+    probePaths: IdeProbePaths,
+    config: DriverConfig,
+    private val pluginsBackup: Path
```
lukaszwawrzyk, line 138: You can omit private val

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -128,3 +130,24 @@ final class InstalledIntelliJ(val root: Path, probePaths: IdeProbePaths, config:
...
+) extends InstalledIntelliJ(root, probePaths, config) {
+  override def cleanup(): Unit = {
+    val pluginsDir = root.resolve("plugins")
+    pluginsDir.delete()
+    Files.move(pluginsBackup, pluginsDir, StandardCopyOption.REPLACE_EXISTING)
```
lukaszwawrzyk, line 143: I think we have a method for that in pathextension, if not, it's best to add one

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
```diff
@@ -37,12 +37,12 @@ trait RunningIntelliJPerSuiteBase {
...
 
   def teardown(): Unit = {
-    try runningIntelliJFixture.foreach(r => AfterTestChecks(baseFixture.factory.config.check, r.probe))
+    try runningIntelliJFixture.foreach(r => AfterTestChecks(baseFixture.intelliJProvider.factory.config.check, r.probe))
     finally {
       try afterAll()
       finally {
         running.foreach(baseFixture.closeIntellij(_))
-        installed.foreach(baseFixture.deleteIntelliJ(_))
+        installed.foreach(baseFixture.cleanupIntelliJ(_))
```
MichalBednarz, line 45: IDK if `cleanupIntelliJ` is a good place for it but I want to have a way to provide my own method that cleans up the instance (maybe something similar to `afterIntelliJInstall`, `afterWorkspaceSetup` or sth?)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -128,3 +130,24 @@ final class InstalledIntelliJ(val root: Path, probePaths: IdeProbePaths, config:
...
+  }
+}
+
+final class DownloadedIntelliJ(
+    val root: Path,
```
lukaszwawrzyk, line 148: not sure if it needs to be val if we delete it from cleanup method

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJFactory.scala
```diff
@@ -1,12 +1,16 @@
...
+import com.google.gson.Gson
 import java.nio.file.{Files, Path}
 import java.util.stream.{Collectors, Stream => JStream}
+
 import org.virtuslab.ideprobe.Extensions._
 import org.virtuslab.ideprobe.IdeProbePaths
 import org.virtuslab.ideprobe.config.{DependenciesConfig, DriverConfig}
 import org.virtuslab.ideprobe.dependencies._
 
+import scala.io.Source
```
lukaszwawrzyk, line 12: it is best not to use this api in prod

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
```diff
@@ -0,0 +1,96 @@
...
+}
+
+case class DefaultIntelliJ(
+    version: IntelliJVersion,
+    plugins: Seq[Plugin],
+    factory: IntelliJFactory
+) extends IntelliJProvider {
+
+  override def setup(): InstalledIntelliJ =
+    factory.create(version, plugins)
```
MichalBednarz, line 74: I would refactor here. `ExistingIntelliJ` in `setup` contains the setup logic but the setup logic of `DefaultIntelliJ` is located in factory. Any idea if there is a better approach?

core/driver/sources/src/test/resources/org/virtuslab/ideprobe/dependencies/product-info.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "IntelliJ IDEA",
+  "version": "2020.2.4",
+  "buildNumber": "202.8194.7",
+  "productCode": "IC",
+  "dataDirectoryName": "IdeaIC2020.2",
+  "svgIconPath": "bin/idea.svg",
+  "launch": [
```
lukaszwawrzyk, line 8: I didn't know about this file. It would be nice to use it in the future to modify how we start ide probe on mac and windows and what default vm options we take. Nice find

kszydlo-vl: All credits go to @MichalBednarz, I'm just feeding on his success.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
```diff
@@ -0,0 +1,76 @@
...
+import org.virtuslab.ideprobe.config.{DependenciesConfig, IdeProbeConfig, IntelliJProvider, IntellijConfig, PathsConfig}
+import org.virtuslab.ideprobe.ide.intellij.IntelliJFactory
+
+import java.nio.file.Path
+
+@RunWith(classOf[JUnit4])
+final class IntelliJCleanupTest {
+  @Test
+  def existingIntelliJShouldNotBeDeletedDuringCleanup: Unit = givenInstalledIntelliJ { installationRoot =>
+      val intelliJProvider = IntelliJProvider.from(
```
lukaszwawrzyk, line 15: it would be best to either create provider in one of two ways:
- instantiate fixture from config in form of string
- create IntelliJProvider using constructor i.e. programatically

Basically there are 2 ways to use probe, either from code using public api and helper `withX` methods or create it from config file/string.

Classes from config package are just implementation detail for parsing the config.

You can notice here also that IdeProbeConfig is too much to pass, as you specify `workspace` here that is unused in the IJProvider

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
```diff
@@ -0,0 +1,76 @@
...
+          paths = PathsConfig()
+        )
+      )
+      val existingInstalledIntelliJ = intelliJProvider.setup()
+
+      //when
+      existingInstalledIntelliJ.cleanup()
+
+      //then
+      assert(installationRoot.toFile.exists(), "The provided IntelliJ instance should exist after cleanup, but it was deleted.")
```
lukaszwawrzyk, line 30: either add exists to PathExtensions or use isFile() or isDirectory()

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+  @Test
+  def existingIntelliJShouldRetainItsOriginalPluginsDuringCleanup: Unit = givenInstalledIntelliJ { installationRoot =>
+    //given a pre-installed IntelliJ and an IntelliJProvider
+    val preInstalledPlugins = installationRoot.resolve("plugins").toFile.list().toSet
+    val pluginsToInstall = Seq( //Plugins choice is arbitrary.
```
lukaszwawrzyk, line 37: this looks unreadable and might change with IJ upgrade. We rather keep our tests using latest version of IJ to make sure it keeps working.
Perhaps you should install some known plugin, like scala plugin in original intellij, then for example install some other plugin or a plugin in different version and then check if scala plugin is back after cleanup.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
```diff
@@ -0,0 +1,76 @@
...
+      Plugin.Versioned("9667", "1.0.298", None),
+      Plugin.Versioned("8327", "2021.1-1.5.9", None)
+    )
+    val intelliJProvider = IntelliJProvider.from(
```
lukaszwawrzyk, line 41: I mentioned earlier to refactor creating this, but if it ends up with so many lines of repetitive code, I'd suggest to extract method.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJFactoryTest.scala
```diff
@@ -0,0 +1,29 @@
...
+  @Test
+  def test: Unit = {
```
lukaszwawrzyk, line 19: perhaps there is a better name for this test

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJFactoryTest.scala
```diff
@@ -0,0 +1,29 @@
...
+    val resourcePath = Paths.get(getClass.getResource(".").getPath)
```
lukaszwawrzyk, line 20: I'd suggest to put the product info in some subdirectory with meaningful name

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
kszydlo-vl: It might be worth creating a separate issue to limit the scope of this one. @lukaszwawrzyk WDYT?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
kszydlo-vl: The two setup processes are fundamentally different - handling existing instances is outside of factory's responsibilities (at least in its current form). The provider makes a decision which approach should be taken, and depending on that either delegates the creation to factory, or just prepares the pre-installed instance for further manipulation.

We could have two specialized "factories" for both default and existing instances, thus eliminating the need for a provider completely. @lukaszwawrzyk WDYT?

build.sbt
```diff
@@ -164,7 +164,7 @@ lazy val scalaProbePlugin =
     .dependsOn(probePlugin, scalaProbeApi)
```
lukaszwawrzyk, line 164: Thanks for implementing this. It looks good overall, I left some comments.

First thing to know is that ide-probe can be configured from config and from code (just create IntelliJFixture from code). IJ fixture can be also tweaked from code after config is loaded with some helper methods. The biggest bug I see is that if someone modifies list of plugins in IJ fixture it won't be reflected, because it is only read from config. Similarly IJ version can't be changed, it is now part of factory. For existing IJ this can't be specified.
I think high level api should be
```scala
fixture.version
IJFixture(intelliJProvider = IntelliJProvider(IntelliJVersion(..., ...)), ...) // apply method that creates default provider, existing workspace is a rare case
IJFixture(intelliJProvider = IntelliJProvider.Existing(path = ..., ...)) // create existing workspace specifying path
fixture.withVersion(IJVersion(..., ...)) // repaces version in default provider if it is default or throws for Existing
```
This is how I imagine code API, config API is already fine.

Also I am concerned about the design of IntelliJ provider and how it contains the factory in both cases.

This looks like a design problem. IntelliJFactory is meant to create InstalledIJ instance that has everything needed to launch it. It was designed to install isolated instance and this is what most of its methods do. Making it a member of existing workspace, that in theory shouldn't need any of these methods and making it a public api (that IJProvider has Factory) is not ideal. 
What I would rather do (I explain the path as if your PR did not exist, you may want to just alter your code to not start from scratch to get to the same result) is:
- Rename IJFactory to IJProvider fo consistency with WorkspaceProvider.
- Split into interface and implementation extracting public api that makes sense for both future implementations to interface. 
-Move whatever needs to be moved to another place, for example version shouldn't be a field of fixture bit be an accessor method and a setter should throw if it can't replace it. Any such changes that make sense.
- Implement the interface for IJProvider.Existing handling anything that needs to be handled.
- If both interface need some common logic, extract it to a separate class, named approprietly after what it is doing. For example PluginInstaller with a clear API.

Also avoid making private methods public just to reuse them or for testing, if these methods are used by other public methods. For example as you did with `version` resolving from file. It would be better to create a separate class, like IJVersionResolver that given a root path of intellij installation can return a version, and test it in an unit test.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
lukaszwawrzyk: config package is for config classes that only exist for mapping purpose

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
lukaszwawrzyk: This is kind of what I meant. IJFactory should be renamed to IJProvider and the other "factory"/provider should handle preparing existing intellij

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ClassFixtures.scala
lukaszwawrzyk: Sure it can be a separate PR. @MichalBednarz what do you need it for? 
We probably will add a method set like: beforeIJClose, beforeIJCleanup, beforeWorkspaceCleanup or afterIJClose, afterIJCleanup, afterWorkspaceCleanup. Not sure which make more sense, this is why I am wondering what do you need it for.

MichalBednarz: @lukaszwawrzyk I want to be able to recreate the instance that is e.g. cleanup caches or totally wipe it out and reinstall.

lukaszwawrzyk: I think we should go with `after` variant maybe.

build.sbt
kszydlo-vl: Will be done in the next commit.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJFactoryTest.scala
kszydlo-vl: Will be done in the next commit.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJFactoryTest.scala
kszydlo-vl: This whole test class will be removed together with factory/provider renaming (and the test will be placed elsewhere) - in the next commit.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
kszydlo-vl: Will be done together with the factory/provider renaming.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
kszydlo-vl: I opted for these plugins as it is highly improbable that they will be pre-installed. I'll fix it together with the factory/provider renaming.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
kszydlo-vl: Done.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
kszydlo-vl: I'll do it in the next commit, together with the Factory renaming.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
kszydlo-vl: Okay, will be done in the next commit.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJFactory.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IntelliJFixture.scala
kszydlo-vl: Fixed.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJFactoryTest.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
kszydlo-vl: Right. Removed.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJFactory.scala
kszydlo-vl: Replaced the implementation in `IntelliJProvider` with a call to this one.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+    path: Path,
+    plugins: Seq[Plugin],
+    factory: IntelliJFactory
+) extends IntelliJProvider {
+  import org.virtuslab.ideprobe.Extensions._
+  override def setup(): InstalledIntelliJ = {
+    val pluginsDir = path.resolve("plugins")
+    val backupDir = Files.createTempDirectory(path, "plugins")
+    pluginsDir.copyDir(backupDir)
+    factory.installPlugins(InternalPlugins.probePluginForIntelliJ(version) +: plugins, path)
```
lukaszwawrzyk, line 28: duplicate logic, surely can be resolved while refactoring the factory.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntelliJProvider.scala
kszydlo-vl: Done.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
kszydlo-vl: Created a fixture (and therefore the provider) from config string.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
kszydlo-vl: Changed to Scala plugin (together with creating the provider from config).

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJCleanupTest.scala
kszydlo-vl: Switched to creating provider from config strings.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJFactoryTest.scala
kszydlo-vl: Renamed the class to `IntelliJVersionResolverTest` and renamed the method.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJFactoryTest.scala
kszydlo-vl: Done.

build.sbt
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntellijConfig.scala
```diff
@@ -1,9 +1,32 @@
...
-import org.virtuslab.ideprobe.dependencies.IntelliJVersion
-import org.virtuslab.ideprobe.dependencies.Plugin
+import java.nio.file.Path
 
-case class IntellijConfig(
-    version: IntelliJVersion = IntelliJVersion.Latest,
-    plugins: Seq[Plugin] = Seq.empty
-)
+import org.virtuslab.ideprobe.ConfigFormat
+import org.virtuslab.ideprobe.dependencies.{IntelliJVersion, Plugin, Resource}
```
MichalBednarz, line 6: remove unused import

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJPaths.scala
```diff
@@ -39,3 +41,31 @@ final class IntelliJPaths(root: Path, headless: Boolean) {
       .makeExecutable()
   }
 }
+
+object IntelliJPaths {
+  def apply(root: Path, headless: Boolean): IntelliJPaths = {
+    val bin = root.resolve("bin")
+    val ideaProperties = Config.fromFile(bin.resolve("idea.properties"))
```
MichalBednarz, line 48: Don't we read those paths from `idea.properties` only to write them again to this file? How about letting the user specify those via config? The other problem that I can spot is that there can be some useful properties other than those paths in `idea.properties` that are going to be wiped out.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+    dependencies: DependencyProvider,
+    override val plugins: Seq[Plugin],
+    override val version: IntelliJVersion,
+    val paths: IdeProbePaths,
+    val config: DriverConfig
+) extends IntelliJProvider {
+  override def withConfig(config: DriverConfig): IntelliJFactory =
+    new IntelliJFactory(dependencies, plugins, version, paths, config)
+
+  override def withPaths(paths: IdeProbePaths): IntelliJFactory =
```
MichalBednarz, line 91: I wonder if adding a way to adjust `IntelliJPaths` in runtime would be a good idea for example with `withIntelliJPaths` method.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IdeaProperties.scala
```diff
@@ -0,0 +1,28 @@
+package org.virtuslab.ideprobe.config
+
+import pureconfig.ConfigReader
+import pureconfig.generic.auto._
+import pureconfig.generic.ProductHint
+
+import java.nio.file.Path
+
+object IdeaProperties {
```
lukaszwawrzyk, line 9: If this is part of configuring ide-probe, I don't think we need to mimic IJ model. If this is just for sake of parsing idea.properties, I think it is a bit of overkill.
Just `Config.fromFile(...).get[String]("java.util.prefs.userRoot")` would do the job.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJPaths.scala
lukaszwawrzyk: Implementation around ij paths and properties is not very straightforward now. 
@MichalBednarz for DownloadedIJ, we don't need to care about location of specific directories like plugins or logs, it could be configurable, currently we can only configure where IJ will be installed along with all of these paths. I'd say it is good enough. As for LocalIJ, location of these paths is already configured in idea.properties, we should just read it and assume that e.g. plugins are wherever this file specifies. In such case we can just read paths from file and use it, no need for further rewrites. If this works for our windows use case, I'd keep it like that for now. However there is more to it, as idea.properties is not the only place that can be used to specify these directories I think. `-Didea.plugin.paths=something-else` would overwrite this, it could be added on invocation or in vmoptions file that can be located in couple of places with different priorities for reading it... But if we can trust the default way for our use cases, let's just do this.

Now to the refactoring:
Let's be consistent now and now make IJPaths a data class and put all setup logic inside companion object factory methods. 
We need 2 ways of creating it:
- read from config file the locations of 4 paths based on `root` and just reference all other paths to hard coded values, like the bin and executable paths.(for local)
- use predefined directories and make sure they are created (for downloaded)

We have additional responsibility here of setting up executable file content and privacy policy. It was here in more accidental way I'd say. It is awkward that IJPaths constructor takes `headless` parameter, it has nothing to do with paths. It would make sense to move the logic that sets up policy and executable to the code that is installing intellij (I know the original bad design was not yours). Replacing executable should happen for both Downloaded and Local, not sure about privacy policy, but I would set it up as well.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+  private def parallel[A](s: Seq[A]) = {
+    s.asJava.parallelStream
+  }
+}
+
+final class ExistingIntelliJ(
+    dependencies: DependencyProvider,
+    path: Path,
+    override val plugins: Seq[Plugin],
+    val paths: IdeProbePaths,
```
lukaszwawrzyk, line 58: inconsistent use of override

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+    new ExistingIntelliJ(dependencies, path, plugins, paths, config)
+
+  override def withPaths(paths: IdeProbePaths): ExistingIntelliJ =
+    new ExistingIntelliJ(dependencies, path, plugins, paths, config)
+
+  override def withPlugins(plugins: Plugin*): ExistingIntelliJ =
+    new ExistingIntelliJ(dependencies, path, this.plugins ++ plugins, paths, config)
+
+  override def setup(): InstalledIntelliJ = {
+    val pluginsDir = path.resolve("plugins")
```
lukaszwawrzyk, line 72: shouldn't this come from IJPaths?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+    installIntelliJ(version, root)
+    installPlugins(dependencies, plugins, root)
+
+    new DownloadedIntelliJ(root, paths, config)
+  }
+
+  private def createInstanceDirectory(version: IntelliJVersion): Path = {
+    val path = paths.instances.createTempDirectory(s"intellij-instance-${version.build}-")
+
+    Files.createDirectories(path)
```
lukaszwawrzyk, line 109: is this needed?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  private def installIntelliJ(version: IntelliJVersion, root: Path): Unit = {
+    println(s"Installing $version")
+    val file = dependencies.fetch(version)
+    file.toArchive.extractTo(root)
+    root.resolve("bin/linux/fsnotifier64").makeExecutable()
```
lukaszwawrzyk, line 116: followup/nice to have - it would be good to traverse bin dir and make everything executable. We have some mess in logs because we don't do this.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
kszydlo-vl: This would be quite easy to implement for downloaded IntelliJ case, but relatively complicated for the existing.

I don't know whether adjusting paths for an existing instance makes much sense, but if it doesn't, we would have a different API for the two `IntelliJProvider` implementations, and again I don't know if the caller should be forced to know which one is being used.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/config/IntellijConfig.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJPaths.scala
kszydlo-vl: `IntelliJPaths` were made a case class, method for reading them from file was implemented, `headless` parameter was moved to `InstalledIntelliJ`. Right now the `LocalIntelliJ` reads the config from `idea.properties`, filling in the gaps with defaults if something is not found, then appends all the settings to the file (in this way we don't overwrite anything we don't recognize). At cleanup the file is restored from backup.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
kszydlo-vl: Done.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
kszydlo-vl: Apparently not - removed.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
kszydlo-vl: Done.

core/api/src/main/scala/org/virtuslab/ideprobe/ProbeExtensions.scala
```diff
@@ -121,6 +123,9 @@ trait ProbeExtensions {
       Files.setPosixFilePermissions(path, attributes)
     }
 
+    def makeExecutableRecursively(): Path =
```
lukaszwawrzyk, line 126: nice

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -128,3 +153,45 @@ final class InstalledIntelliJ(val root: Path, probePaths: IdeProbePaths, config:
...
+    pluginsBackup: Path
+) extends InstalledIntelliJ(root, probePaths, config) {
+  override protected val ideaProperties: Path = root.resolve("bin").resolve("idea.properties")
+
+  private val ideaPropertiesBackup: Option[Path] = if (ideaProperties.isFile) {
+    val tempPath = Files.createTempFile(root, "idea.properties", ".backup")
+    Some(ideaProperties.copyTo(tempPath))
+  } else None
+
+  Files.write(ideaProperties, ideaPropertiesContent.getBytes(), StandardOpenOption.APPEND)
```
lukaszwawrzyk, line 171: Don't we have something like `path.write(content, append = true)` or `path.append(content)`? If not, it is worth adding.

As fo the logic, if we *append* our config won't we override existing paths?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -128,3 +153,45 @@ final class InstalledIntelliJ(val root: Path, probePaths: IdeProbePaths, config:
...
+    cleanupIdeaProperties()
+    val pluginsDir = root.resolve("plugins")
+    pluginsDir.delete()
+    pluginsBackup.moveTo(pluginsDir)
+  }
+
+  private def cleanupIdeaProperties(): Unit = ideaPropertiesBackup.foreach { backup =>
+    ideaProperties.delete()
+    backup.copyTo(ideaProperties)
+    backup.delete()
```
lukaszwawrzyk, line 183: perhaps just moveTo would do it

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJPaths.scala
```diff
@@ -1,41 +1,43 @@
...
 
-      s"""|#!/bin/sh
-          |$command "$$@"
-          |""".stripMargin
+    val configPath = ideaProperties.get[Path]("idea.config.path").getOrElse(root.createDirectory("config"))
+    val systemPath = ideaProperties.get[Path]("idea.system.path").getOrElse(root.createDirectory("system"))
+    val pluginsPath = ideaProperties.get[Path]("idea.plugins.path").getOrElse(root.createDirectory("plugins"))
+    val logsPath = ideaProperties.get[Path]("idea.log.path").getOrElse(root.createDirectory("logs"))
+    val userPrefsPath = ideaProperties.get[Path]("java.util.prefs.userRoot").getOrElse {
```
lukaszwawrzyk, line 28: @MichalBednarz is it safe to assume that existing ij will have the privacy policy agreement set up? Or should we set it up for both variants?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
```diff
@@ -0,0 +1,183 @@
...
+    new IntelliJFactory(dependencies, plugins, version, paths, config)
+
+  override def withPlugins(plugins: Plugin*): IntelliJProvider =
+    new IntelliJFactory(dependencies, plugins = this.plugins ++ plugins, version, paths, config)
+
+  override def setup(): InstalledIntelliJ = {
+    val root = createInstanceDirectory(version)
+
+    val intelliJPaths: IntelliJPaths =
+      IntelliJPaths(
```
lukaszwawrzyk, line 107: oh, like this...
I thought it would be another static method for `IntelliJPaths` that would hide this details, and possibly share some code with the other method. For example 
`IntelliJPaths.fromExistingInstance(root)` or `IntelliJPaths.fromExistingPreferences(root.resolve("bin/idea.prefs"))`
and
`IntelliJPaths.default(root)`
or something like this

kszydlo-vl: Done.

core/api/src/main/scala/org/virtuslab/ideprobe/ProbeExtensions.scala
kszydlo-vl: Thanks!

lukaszwawrzyk (approved):

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJPaths.scala
MichalBednarz: Setting privacy policy agreement is being taken care of in the case that I'm going to use this feature for. We can skip it for existing instance for now.

kszydlo-vl: Done - no longer trying to install the policy.

MichalBednarz:
I tested it e2e and everything seems to work fine. The only problem that I can spot is when execution fails for example because of some permission issues. Then the cleanup logic doesn't apply and next probe run fails on trying to install the plugins that are already there.

PR #132, opened 2021-06-07
Special logger for JsonRpcEndpoint

### LINKED ISSUE #121: Minor: `request[backgroundTasks]: {"data":null} response: {"data":[]}` make a lot of spam
Maybe this could be skipped in the logs when data is empty?

kszydlo-vl:
A RequestResponseLogger capable of buffering and counting repeated requests and
responses was created and plugged into JsonRpcEndpoint to reduce noise
in the logs.

Fixes #121

### DIFF
--- a/core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/JsonRpcEndpoint.scala
+++ b/core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/JsonRpcEndpoint.scala
@@ -9,7 +9,7 @@ import scala.concurrent.Future
 import scala.reflect.ClassTag
 
 trait JsonRpcEndpoint extends AutoCloseable {
-  private val logger: RequestResponseLogger = new RequestResponseLogger
+  private val logger = new RequestResponseLogger
   protected def connection: JsonRpcConnection
 
   implicit protected def ec: ExecutionContext
--- a/core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/JsonRpcEndpoint.scala
+++ b/core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/JsonRpcEndpoint.scala
@@ -2,11 +2,14 @@ package org.virtuslab.ideprobe.jsonrpc
 
 import org.virtuslab.ideprobe.Close
 import org.virtuslab.ideprobe.jsonrpc.JsonRpc._
+import org.virtuslab.ideprobe.jsonrpc.logging.RequestResponseLogger
+
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.reflect.ClassTag
 
 trait JsonRpcEndpoint extends AutoCloseable {
+  private val logger: RequestResponseLogger = new RequestResponseLogger
   protected def connection: JsonRpcConnection
 
   implicit protected def ec: ExecutionContext
@@ -17,12 +20,12 @@ trait JsonRpcEndpoint extends AutoCloseable {
     Future(method.encode(parameters)).flatMap { json =>
       method match {
         case Method.Notification(name) =>
-          println(s"notification: ${name}")
-          println(s"value: ${json}")
+          println(s"notification: $name")
+          println(s"value: $json")
           connection.sendNotification(name, json)
           Future.unit.asInstanceOf[Future[B]]
         case Method.Request(name) =>
-          println(s"request[$name]: $json")
+          logger.logRequest(s"request[$name]: $json")
           connection
             .sendRequest(name, json)
             .flatMap {
@@ -34,10 +37,10 @@ trait JsonRpcEndpoint extends AutoCloseable {
                   new RemoteException(cause)
                 }
 
-                println(s"response: ${error.message}")
+                logger.logResponse(s"response: ${error.message}")
                 Future.failed(exception)
               case response =>
-                println(s"response: ${response.result}")
+                logger.logResponse(s"response: ${response.result}")
                 Future(method.decode(response.result))
             }
       }
--- a/core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
+++ b/core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
@@ -1,35 +1,38 @@
 package org.virtuslab.ideprobe.jsonrpc.logging
 
-import java.time.ZonedDateTime
-import java.util.concurrent.{ScheduledFuture, ScheduledThreadPoolExecutor, TimeUnit}
+import java.time.Instant
+import java.util.concurrent.{ScheduledFuture, ScheduledThreadPoolExecutor}
+import scala.concurrent.duration.{Duration, SECONDS}
 
 class RequestResponseLogger {
 
   private val ex = new ScheduledThreadPoolExecutor(1)
-  private val ticker: Runnable = () => flush()
-  private var task: Option[ScheduledFuture[_]] = None
+  private val flushTask: Runnable = () => flush()
+  private var scheduledTask: Option[ScheduledFuture[_]] = None
+
+  private val flushTimeout = Duration(1, SECONDS)
 
   private sealed trait LogEntry
   private case class Request(message: String) extends LogEntry
   private case class Response(message: String) extends LogEntry
   private case class RequestAndResponse(request: String, response: String) extends LogEntry
 
-  private var buffer: List[(LogEntry, ZonedDateTime)] = List.empty
+  private var buffer: List[(LogEntry, Instant)] = List.empty
 
   def logRequest(request: String): Unit = buffer.synchronized {
-    task.foreach(_.cancel(true))
+    scheduledTask.foreach(_.cancel(true))
     buffer match {
       case (head @ (RequestAndResponse(`request`, _), _)) :: tail =>
-        buffer = (Request(request), ZonedDateTime.now()) :: head :: tail
+        buffer = (Request(request), Instant.now()) :: head :: tail
       case _ =>
         flush()
-        buffer = (Request(request), ZonedDateTime.now()) :: Nil
+        buffer = (Request(request), Instant.now()) :: Nil
     }
-    task = Option(ex.schedule(ticker, 1, TimeUnit.SECONDS))
+    scheduledTask = Option(ex.schedule(flushTask, flushTimeout.length, flushTimeout.unit))
   }
 
   def logResponse(response: String): Unit = buffer.synchronized {
-    task.foreach(_.cancel(true))
+    scheduledTask.foreach(_.cancel(true))
     buffer match {
       case (Request(request), timestamp) :: Nil =>
         buffer = (RequestAndResponse(request, response), timestamp) :: Nil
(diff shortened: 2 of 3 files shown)


core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
```diff
@@ -0,0 +1,76 @@
...
+  private val ex = new ScheduledThreadPoolExecutor(1)
+  private val ticker: Runnable = () => flush()
+  private var task: Option[ScheduledFuture[_]] = None
+
+  private sealed trait LogEntry
+  private case class Request(message: String) extends LogEntry
+  private case class Response(message: String) extends LogEntry
+  private case class RequestAndResponse(request: String, response: String) extends LogEntry
+
+  private var buffer: List[(LogEntry, ZonedDateTime)] = List.empty
```
lukaszwawrzyk, line 17: perhaps `Instant` is more suitable class for this use case?

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
```diff
@@ -0,0 +1,76 @@
...
+  def logRequest(request: String): Unit = buffer.synchronized {
+    task.foreach(_.cancel(true))
+    buffer match {
+      case (head @ (RequestAndResponse(`request`, _), _)) :: tail =>
+        buffer = (Request(request), ZonedDateTime.now()) :: head :: tail
+      case _ =>
+        flush()
+        buffer = (Request(request), ZonedDateTime.now()) :: Nil
+    }
+    task = Option(ex.schedule(ticker, 1, TimeUnit.SECONDS))
```
lukaszwawrzyk, line 28: let's extract this 1 second to a val of type scala.concurrent.Duration. We might further put it in configuration

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/JsonRpcEndpoint.scala
```diff
@@ -2,11 +2,14 @@ package org.virtuslab.ideprobe.jsonrpc
...
 import org.virtuslab.ideprobe.Close
 import org.virtuslab.ideprobe.jsonrpc.JsonRpc._
+import org.virtuslab.ideprobe.jsonrpc.logging.RequestResponseLogger
+
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.reflect.ClassTag
 
 trait JsonRpcEndpoint extends AutoCloseable {
+  private val logger: RequestResponseLogger = new RequestResponseLogger
```
lukaszwawrzyk, line 12: no need for type on val

kszydlo-vl: Done.

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
kszydlo-vl: Done.

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
kszydlo-vl: Done.

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
```diff
@@ -0,0 +1,76 @@
...
+      case _ =>
+        buffer = (Response(response), ZonedDateTime.now()) :: buffer
+        flush()
+    }
+    task = Option(ex.schedule(ticker, 1, TimeUnit.SECONDS))
+  }
+
+  private def flush(): Unit = buffer.synchronized {
+    val now = ZonedDateTime.now()
+    val firstMessageTimestamp = buffer.reverse.headOption.map(_._2).getOrElse(now)
```
lukaszwawrzyk, line 54: wouldn't this mean an empty list? How about early return? If list is empty - do nothing, else take last element, and don't bother handling options. Currently it feels like this `now` in firstMessageTimestamp is used for anything

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
```diff
@@ -0,0 +1,76 @@
...
+                 |}""".stripMargin)
+      case (Request(request), _) => //Here and below the count will always be 1 - see the buffering mechanism.
+        println(request)
+      case (Response(response), _) =>
+        println(response)
+    }
+    buffer = Nil
+  }
+
+  private def collectCountingSubsequent[X](xs: List[X]): List[(X, Int)] =
```
lukaszwawrzyk, line 72: Quite complex piece of logic (like whole class actually, but the problem itself is quite complex).  Reversing is a bit confusing, would foldRight help? (not sure).

I also found something, perhaps simpler, though recursive

```
def collectCountingSubsequent[A](xs:List[A]): List[(A,Int)] = {
   xs match {
    case Nil => Nil
    case head :: _ =>
      val (repeated, rest) = xs.span(_ == head)
      (head, repeated.length) :: collectCountingSubsequent(rest)
  }
}
```

core/api/src/test/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLoggerTest.scala
```diff
@@ -0,0 +1,139 @@
+package org.virtuslab.ideprobe.jsonrpc.logging
+
+import org.junit.{Assert, Test}
+
+import java.io.ByteArrayOutputStream
+
+class RequestResponseLoggerTest {
```
lukaszwawrzyk, line 7: nice test coverage

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
kszydlo-vl: Yeah, it is in fact unused when the list is empty. Fixed.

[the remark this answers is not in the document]
kszydlo-vl: Done.

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
kszydlo-vl: Thanks for the code snippet - I made it tail-recursive.

lukaszwawrzyk:
```
Repeated 11 times over 0 seconds: {
  request[backgroundTasks]: {"data":null}
  response: {"data":["ProgressIndicator 441017100: running=true; canceled=false"]}
}
```
Found this in your log, perhaps we should have special variant of message for 0 seconds and 1 second

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
```diff
@@ -0,0 +1,93 @@
...
+  private val flushTimeout = Duration(1, SECONDS)
+
+  private sealed trait LogEntry
+  private case class Request(message: String) extends LogEntry
+  private case class Response(message: String) extends LogEntry
+  private case class RequestAndResponse(request: String, response: String) extends LogEntry
+
+  private var buffer: List[(LogEntry, Instant)] = List.empty
+
+  def logRequest(request: String): Unit = buffer.synchronized {
```
lukaszwawrzyk, line 23: Synchronization is rather not needed at all, but if you synchronize on something, it would make more sense to synchronize on `this`. The buffer is changing.

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
```diff
@@ -0,0 +1,93 @@
...
+                     |  $request
+                     |  $response
+                     |}""".stripMargin)
+        case (Request(request), _) => //Here and below the count will always be 1 - see the buffering mechanism.
+          println(request)
+        case (Response(response), _) =>
+          println(response)
+      }
+      buffer = Nil
+    } else ()
```
lukaszwawrzyk, line 72: no need for `else ()`, compiler will add this for you

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
```diff
@@ -0,0 +1,93 @@
...
+  private def flush(): Unit = buffer.synchronized {
+    if (buffer.nonEmpty) {
+      val now = Instant.now()
+      val firstMessageTimestamp = buffer.last._2
+      collectCountingSubsequent(buffer.map(_._1).reverse).foreach {
+        case (RequestAndResponse(request, response), 1) =>
+          println(request)
+          println(response)
+        case (RequestAndResponse(request, response), count) =>
+          println(s"""Repeated $count times ${formatSeconds(java.time.temporal.ChronoUnit.SECONDS.between(firstMessageTimestamp, now))}: {
```
lukaszwawrzyk, line 62: import ChronoUnit at the top of the file

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
kszydlo-vl: Kept the synchronization to avoid potential clashes with scheduled flushing. Now syncing on `this`.

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
kszydlo-vl: Done.

core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/logging/RequestResponseLogger.scala
kszydlo-vl: Done.

lukaszwawrzyk:
Good, to merge I think.
 I looked into CI logs, and I have one last reflection. We group it with `C` syntax. WDYT about `Python`? Maybe it would be more clear:

```
Repeated 10 times over less than second {
  request[a]: {}
  response: []
}
request[a]: {}
response: []
Repeated 10 times over less than second {
  request[a]: {}
  response: []
}
request[a]: {}
response: []
```
vs
```
Repeated 10 times over less than second:
  request[a]: {}
  response: []
request[a]: {}
response: []
Repeated 10 times over less than second:
  request[a]: {}
  response: []
request[a]: {}
response: []
```

I am not sure myself...

PawelLipski:
Wow @kszydlo-vl you're contributing here as well? 😯

PawelLipski:
Actually, out of curiosity, when is this change going to be released?

lukaszwawrzyk:
Hopefully tomorrow in maven central as 0.9.0

PR #134, opened 2021-06-07
Performing cleanup even after IDE fails to start

### LINKED ISSUE #133: SingleRunIntelliJ does not perform cleanup if starting the IDE fails.
If for some reason the `idea.sh` file is missing, or has incorrect permissions, or simply fails to execute, the IDE's instance directory and its workspace don't get deleted.

kszydlo-vl:
Closes #133

### DIFF
--- a/core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/SingleRunFixtureTest.scala
+++ b/core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/SingleRunFixtureTest.scala
@@ -49,4 +49,37 @@ final class SingleRunFixtureTest extends IdeProbeFixture with WorkspaceFixture w
     assertFalse("Workspace was not removed", Files.exists(workspace))
     assertFalse("IDE home was not removed", Files.exists(ideHome))
   }
+
+  @Test
+  def removesDirectoriesEvenAfterFailureToRunIntelliJ(): Unit = {
+    val intelliJFixture = IntelliJFixture().withAfterIntelliJInstall((_, intellij) =>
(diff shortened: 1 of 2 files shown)


core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/SingleRunFixtureTest.scala
```diff
@@ -49,4 +49,40 @@ final class SingleRunFixtureTest extends IdeProbeFixture with WorkspaceFixture w
...
+  @Test
+  def removesDirectoriesEvenAfterFailureToRunIntelliJ(): Unit = {
+    val intelliJFixture = IntelliJFixture(
+      version = IntelliJVersion.Latest,
+      plugins = List()
+    ).withAfterIntelliJInstall((_, intellij) =>
+      Files.delete(intellij.root.resolve("bin").resolve("idea.sh")) //To prevent the IDE from launching.
+    )
+
+    val instancesBefore = Option(intelliJFixture.factory.paths.instances.toFile.list().toSet).getOrElse(Set.empty)
```
lukaszwawrzyk, line 62: It would be good to have some helper method for these 4 long chains of calls in this test.
There should be PathExtension util for listing files/directories, if not, it would be good to add. 
Also, why Option? How could `toSet` return null?

kszydlo-vl: Indeed there is a PathExtension method for that. Using it now.

The Option here was to avoid getting NPE, since `File.list` method can return null - but I misplaced the `toSet` call. Anyway, it's probably unnecessary as the directories should be created at this point.

lukaszwawrzyk:
Good to merge after resolving conflicts

PR #182, opened 2022-05-16
Fallback to other resolvers if file resource is missing

wiacekm:
When locally stored, unzipped resource (ex. Intellij) was missing it was `NoSuchFileException` was thrown. Now it is handled correctly.


PawelLipski (approved):
Tested locally, works fine now 👌🏻

Issue #23, opened 2020-07-31
Fetching remote resources (and cloning git repositories) could use a progress indicator


bednam:
The progress indicator in the console would be helpful while fetching SDK or remote repo.

PR #198, opened 2022-06-21
use progress bar

wiacekm:


lukaszwawrzyk:
Thank you!

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/WorkspaceProvider.scala
```diff
@@ -114,11 +114,10 @@ object WorkspaceTemplate {
...
   case class FromGit(repository: String, ref: Option[String]) extends WorkspaceTemplate {
     override def setupIn(workspace: Path): Unit = {
-      val cloned = Shell.run("git", "clone", repository, workspace.toString)
-      if (cloned.exitCode != 0) throw new IllegalStateException(s"Could not clone git $repository")
-      ref.foreach { ref =>
-        val checkout = Shell.run(in = workspace, "git", "checkout", ref)
-        if (checkout.exitCode != 0) throw new IllegalStateException(s"Could not checkout $ref in $repository")
+      import org.virtuslab.ideprobe.dependencies.git.GitHandler._
+      val git = repository.clone(workspace)
+      ref.foreach{
```
lukaszwawrzyk, line 119: Maybe the api for git can be cleaned up a bit without many implicit conversions:
let's have a factory method to create git repo by cloning it to some location and such repo will have the necessary methods.
```
val repo = GitRepository.clone(repository, workspace)
ref.foreach(repo.checkout)
```

The GitRepository already exists in dependencies though and it looks similar to what I suggested. 
I guess it could stay as is, but it would be ideal to refactor all git interactions in this repo to rely on one common api.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/ResourceProvider.scala
```diff
@@ -42,12 +43,10 @@ object ResourceProvider {
...
       val cachedResource = cached(uri)
       if (!cachedResource.isFile) {
-        val stream = createStream
         println(message(cachedResource))
-        Files
-          .createTempFile("cached-resource", "-tmp")
-          .append(stream)
-          .moveTo(cachedResource)
+        val tempFile: Path = Files.createTempDirectory("cached-resource")
+        val downloader = FileDownloader(tempFile).download(uri.toURL)
```
lukaszwawrzyk, line 48: the val doesn't contain downloader but the downloaded file path

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/git/GitHandler.scala
```diff
@@ -0,0 +1,97 @@
...
+  implicit class GitRepository(repository: String) {
+    def clone(workspace: Path): Git = {
+      val command: CloneCommand = new CloneCommand()
+        .setURI(repository)
+        .setDirectory(workspace.toFile)
+        .setCloneAllBranches(true)
+        .setMirror(false)
+        .setBare(false)
+        .setProgressMonitor(new CustomProgressMonitor(repository))
+      val git = Try(command.call())
```
lukaszwawrzyk, line 23: I think that the standard try catch is simpler in this case (and couple other in this PR)
```
try command.call()
catch {
  case e: Exception => throw new IllegalStateException(s"Could not clone git $repository: ${e.getMessage}")
}
```

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/git/GitHandler.scala
```diff
@@ -0,0 +1,97 @@
...
+      res
+    }
+  }
+
+  private class CustomProgressMonitor(repository: String) extends ProgressMonitor {
+
+    private val totalWork: AtomicInteger = new AtomicInteger()
+    private val completion: AtomicInteger = new AtomicInteger()
+    private val task: AtomicReference[String] = new AtomicReference()
+    private val cancelled: AtomicBoolean = new AtomicBoolean(false)
```
lukaszwawrzyk, line 58: seems unused (never set)

lukaszwawrzyk (approved):

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/WorkspaceProvider.scala
lukaszwawrzyk: Well, some proper refactoring would be to:
have some (new) GitRepository class with git operations like checkout, push, etc. (whatever we need)
This class can have static factory methods like GitRepository.clone(from, branch?), GitRepository.existing(path) (and maybe some useful method cloneAndCheckout.
Only this class would contain any git logic and would use jGit under the hood, report progress etc.
Any other classes that represent path to git repository and a branch that we use to setup pants or workspace would be plain data classes. These data classes could be converted to git repository, either explicitly: GitRepository.cloneAndCheckout(repo.path, repo.ref) or with some helper methods.

wiacekm:
closes #23

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/WorkspaceProvider.scala
wiacekm: @lukaszwawrzyk can we mark it as solved?

PR #204, opened 2022-06-24
Add 'refresh all' endpoint

tpasternak:


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ProbeDriver.scala
```diff
@@ -86,6 +86,10 @@ class ProbeDriver(
     }
   }
 
+  def refreshAllExternalProjectsAsync(project: ProjectRef): Unit = {
```
lukaszwawrzyk, line 89: add = ProjectRef.Default

tpasternak: Done with method overloading (to make it callable from non-scala client code)

lukaszwawrzyk (approved):

PR #224, opened 2022-07-07
Update remote-robot and sbt-idea-plugin

PawelLipski:


build.sbt
```diff
@@ -4,6 +4,8 @@ val scala212 = "2.12.16"
 val scala213 = "2.13.8"
 val crossScalaVersions = List(scala212, scala213)
 
+Global / onChangedBuildSource := ReloadOnSourceChanges
```
PawelLipski, line 7: Unrel, pretty useful: no need to `reload` on every change in sbt config

build.sbt
```diff
@@ -258,7 +260,7 @@ lazy val bazelProbeDriver =
   module(id = "bazel-probe-driver", path = "extensions/bazel/driver")
     .settings(
       name := "bazel-probe-driver",
-      libraryDependencies += "commons-codec" % "commons-codec" % "1.15"
+      libraryDependencies ++= Seq(Dependencies.commonsCodec, Dependencies.commonsIO)
```
PawelLipski, line 263: Added commons-io, which was available indirectly, as a transitive dependency of remote-robot (via logback-core). Once remote-robot removed its dependency on logback (https://github.com/JetBrains/intellij-ui-test-robot/issues/191), now commons-io wasn't available to bazel-probe-driver.

PR #235, opened 2022-07-27
Allow shortening command line when running tests

lukaszwawrzyk:


PawelLipski:
@lukaszwawrzyk out of curiosity... what's the effect of this PR in practice? 🤔 as in, what command line can be shortened from what to what?

lukaszwawrzyk:
Mostly the classpath. It can be longer that system allows when running a command. 
image

PR #241, opened 2022-08-08
Fix most of the compilation warnings

### LINKED ISSUE #223: Fix warnings about unused vars/params
```
[warn] /home/plipski/ide-probe/core/driver/test-plugin/src/main/scala/org/virtuslab/ideprobe/test/OpenWindowAction.scala:11:9: local val p in method actionPerformed is never used
[warn]     val p = e.getProject
[warn]         ^
[warn] one warning found
[info]   Compilation completed in 28.58s.
[warn] /home/plipski/ide-probe/core/driver/test-plugin/src/main/scala/org/virtuslab/ideprobe/test/OpenWindowAction.scala:11:9: local val p in method actionPerformed is never used
[warn]     val p = e.getProject
[warn]         ^
[warn] /home/plipski/ide-probe/core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/JsonRpc.scala:72:42: parameter value evidence$1 in class Method is never used
[warn]   sealed abstract class Method[Parameters: ClassTag: ConfigConvert, Result: ClassTag: ConfigConvert] {
[warn]                                          ^
[warn] one warning found
[warn] /home/plipski/ide-probe/core/api/src/main/scala/org/virtuslab/ideprobe/jsonrpc/JsonRpcEndpoint.scala:19:17: parameter value evidence$1 in method sendRequest is never used
[warn]   protected def sendRequest[A: ClassTag, B: ClassTag](method: Method[A, B], parameters: A): Future[B] = {
[warn]                 ^
```

In fact, I'd elevate all warnings to errors (`-Xfatal-warnings`).

image

LukaszKontowski:
Closes #223

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -70,7 +70,8 @@ lazy val ci = project("ci", "ci", publish = false)
 lazy val api = project("api", "core/api", publish = true)
   .settings(
     libraryDependencies ++= Dependencies.pureConfig,
-    libraryDependencies += Dependencies.gson
+    libraryDependencies += Dependencies.gson,
+    libraryDependencies += Dependencies.scalaCollectionCompat
   )
   .cross
 
@@ -312,7 +313,7 @@ lazy val benchmarks212 = benchmarks(scala212)
 def project(id: String, path: String, publish: Boolean): Project = {
   Project(id, sbt.file(path))
     .settings(scalafmtOnCompile := true)
-    .settings(scalacOptions ++= Seq("-Xlint:unused", "-Ywarn-unused"))
+    .settings(scalacOptions ++= Seq("-Xlint:unused", "-Ywarn-unused", "-deprecation", "-Ywarn-macros:after"))
     .settings(
       (Keys.publish / skip) := !publish,
       libraryDependencies ++= Dependencies.junit
--- a/build.sbt
+++ b/build.sbt
@@ -313,7 +313,15 @@ lazy val benchmarks212 = benchmarks(scala212)
 def project(id: String, path: String, publish: Boolean): Project = {
   Project(id, sbt.file(path))
     .settings(scalafmtOnCompile := true)
-    .settings(scalacOptions ++= Seq("-Xlint:unused", "-Ywarn-unused", "-deprecation", "-Ywarn-macros:after"))
+    .settings(
+      scalacOptions ++= Seq(
+        "-Xlint:unused",
+        "-Ywarn-unused",
+        "-deprecation",
+        "-Ywarn-macros:after",
+        "-Wconf:msg=@nowarn annotation does not suppress any warnings:s"
+      )
+    ) // due to false positive for IdeaLogInterceptor.scala
     .settings(
       (Keys.publish / skip) := !publish,
       libraryDependencies ++= Dependencies.junit
--- a/build.sbt
+++ b/build.sbt
@@ -319,9 +319,9 @@ def project(id: String, path: String, publish: Boolean): Project = {
         "-Ywarn-unused",
         "-deprecation",
         "-Ywarn-macros:after",
-        "-Wconf:msg=@nowarn annotation does not suppress any warnings:s"
+        "-Wconf:msg=@nowarn annotation does not suppress any warnings:s" // due to false positive for IdeaLogInterceptor.scala
       )
-    ) // due to false positive for IdeaLogInterceptor.scala
+    )
     .settings(
       (Keys.publish / skip) := !publish,
       libraryDependencies ++= Dependencies.junit
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/Shell.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/Shell.scala
@@ -13,7 +13,6 @@ import scala.concurrent.duration.Duration
 
 import com.zaxxer.nuprocess.NuAbstractProcessHandler
 import com.zaxxer.nuprocess.NuProcess
-import com.zaxxer.nuprocess.NuProcessBuilder
 import com.zaxxer.nuprocess.NuProcessHandler
 
 object Shell extends BaseShell
@@ -118,7 +117,7 @@ class BaseShell {
     import com.zaxxer.nuprocess._
     val builder = new NuProcessBuilder(command: _*)
     builder.setCwd(in)
-    customizeBuilder(builder)
+    customizeBuilder()
     env.foreach(e => builder.environment().put(e._1, e._2))
     val finished = Promise[CommandResult]()
     val outputCollector = new ProcessOutputCollector
@@ -172,5 +171,5 @@ class BaseShell {
     Seq(outputForwarder)
   }
 
-  protected def customizeBuilder(builder: NuProcessBuilder): Unit = ()
+  protected def customizeBuilder(): Unit = ()
 }
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/Shell.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/Shell.scala
@@ -6,6 +6,8 @@ import java.nio.channels.Channels
 import java.nio.charset.StandardCharsets
 import java.nio.file.Path
 
+import scala.annotation.nowarn
+import scala.collection.mutable
 import scala.concurrent.Await
 import scala.concurrent.Future
 import scala.concurrent.Promise
@@ -13,6 +15,7 @@ import scala.concurrent.duration.Duration
 
 import com.zaxxer.nuprocess.NuAbstractProcessHandler
 import com.zaxxer.nuprocess.NuProcess
+import com.zaxxer.nuprocess.NuProcessBuilder
 import com.zaxxer.nuprocess.NuProcessHandler
 
 object Shell extends BaseShell
@@ -73,8 +76,8 @@ class BaseShell {
   }
 
   class ProcessOutputCollector extends NuAbstractProcessHandler {
-    private val outputBuilder = new StringBuilder
-    private val errorBuilder = new StringBuilder
+    private val outputBuilder = new mutable.StringBuilder
+    private val errorBuilder = new mutable.StringBuilder
 
     override def onStdout(buffer: ByteBuffer, closed: Boolean): Unit = {
       outputBuilder.append(StandardCharsets.UTF_8.decode(buffer))
@@ -117,7 +120,7 @@ class BaseShell {
     import com.zaxxer.nuprocess._
     val builder = new NuProcessBuilder(command: _*)
     builder.setCwd(in)
-    customizeBuilder()
+    customizeBuilder(builder)
     env.foreach(e => builder.environment().put(e._1, e._2))
     val finished = Promise[CommandResult]()
     val outputCollector = new ProcessOutputCollector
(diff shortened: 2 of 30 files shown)


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/Shell.scala
```diff
@@ -172,5 +171,5 @@ class BaseShell {
     Seq(outputForwarder)
   }
 
-  protected def customizeBuilder(builder: NuProcessBuilder): Unit = ()
```
lukaszwawrzyk, line 175: how are inheritors supposed to customizeBuilder if you take away the builder?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/Shell.scala
```diff
@@ -118,7 +117,7 @@ class BaseShell {
     import com.zaxxer.nuprocess._
     val builder = new NuProcessBuilder(command: _*)
     builder.setCwd(in)
-    customizeBuilder(builder)
+    customizeBuilder()
```
PawelLipski, line 120: :suspicious:

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/Shell.scala
```diff
@@ -172,5 +171,5 @@ class BaseShell {
     Seq(outputForwarder)
   }
 
-  protected def customizeBuilder(builder: NuProcessBuilder): Unit = ()
+  protected def customizeBuilder(): Unit = ()
```
PawelLipski, line 174: Oh! indeed, this looks like a method meant for extension

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
```diff
@@ -7,19 +7,19 @@ import java.nio.file.Files
...
 import org.virtuslab.ideprobe.Extensions._
 
+import scala.annotation.nowarn
+
 sealed trait Resource
 
 object Resource extends ConfigFormat {
   implicit val resourceConfigReader: ConfigReader[Resource] = ConfigReader[String].map(from)
 
+  @nowarn // as match is not exhaustive - case class Jar is not handled
```
PawelLipski, line 22: Hmmm actually... maybe it should be handled here? 🤔 or if we know from somewhere else that Jar will never be passed here (doubt), then a separate trait should be extracted? pls open an issue (as I suppose it's not trivial enough to be fixed on this PR)

core/probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/ExpandMacro.scala
```diff
@@ -30,7 +31,7 @@ object ExpandMacro {
     val data = new THashMap[String, Object]()
     data.put(CommonDataKeys.PROJECT.getName, fileRef.project)
     data.put(CommonDataKeys.VIRTUAL_FILE.getName, file)
-    data.put(PlatformDataKeys.PROJECT_FILE_DIRECTORY.getName, project.getBaseDir)
+    data.put(PlatformDataKeys.PROJECT_FILE_DIRECTORY.getName, ProjectUtil.guessProjectDir(project))
```
PawelLipski, line 34: Huh is this equivalent to `getBaseDir`? 🤔

PawelLipski: I suppose that was what the deprecation warning pointed to?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/Shell.scala
LukaszKontowski: You're right - it would make no sense. I just did not find any implementation of this method in the source code and wanted to get rid of the warning

core/probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/IntelliJApi.scala
```diff
@@ -81,13 +80,13 @@ trait IntelliJApi {
...
 
-    def field[B: ClassTag](name: String): B = {
+    def field[B](name: String): B = {
       val field = getField(obj.getClass, "my" + name.capitalize)
       using(field)(_.get(obj).asInstanceOf[B])
     }
 
     private def using[B <: AccessibleObject, C](accessible: B)(f: B => C): C = {
-      val isAccessible = accessible.isAccessible
+      val isAccessible = accessible.canAccess(null) // isAccessible
```
PawelLipski, line 89: Whoops deprecated?

Pls clarify in the comment what you mean by `isAccessible` (it won't be clear soon)

build.sbt
```diff
@@ -312,7 +313,7 @@ lazy val benchmarks212 = benchmarks(scala212)
 def project(id: String, path: String, publish: Boolean): Project = {
   Project(id, sbt.file(path))
     .settings(scalafmtOnCompile := true)
-    .settings(scalacOptions ++= Seq("-Xlint:unused", "-Ywarn-unused"))
+    .settings(scalacOptions ++= Seq("-Xlint:unused", "-Ywarn-unused", "-deprecation", "-Ywarn-macros:after"))
```
PawelLipski, line 316: `-Xfatal-warnings` FTW

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/Shell.scala
PawelLipski: Huh the other problem might be that... will this builder ever be customized indeed? 😅 @lukaszwawrzyk was the customization added just in case?

core/probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/IntelliJApi.scala
LukaszKontowski: just forgot to remove the `// isAccessible` - will do that in next commit. `isAccessible` is deprecated. Moreover, I will have to double check if `canAccess(null)` should be used - or `canAccess(obj: Object)` with some real object... I am not sure yet after quick read of the source code of this method.

core/probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/ExpandMacro.scala
LukaszKontowski: It should be, if I get the javadoc in source code right...
```
/**
 *  Tries to guess the "main project directory" of the project.
 *
 *  There is no strict definition of what is a project directory, since a project can contain multiple modules located in different places,
 *  and the `.idea` directory can be located elsewhere (making the popular [Project.getBaseDir] method not applicable to get the "project
 *  directory"). This method should be preferred, although it can't provide perfect accuracy either.
 */
fun Project.guessProjectDir() : VirtualFile? {
  if (isDefault) {
    return null
  }

  val modules = ModuleManager.getInstance(this).modules
  val module = if (modules.size == 1) modules.first() else modules.firstOrNull { it.name == this.name }
  module?.guessModuleDir()?.let { return it }
  return LocalFileSystem.getInstance().findFileByPath(basePath!!)
}
```

PawelLipski: Actually, it's [KDoc](https://kotlinlang.org/docs/kotlin-doc.html) in this case ;)

PawelLipski: But javadoc/KDoc/scaladoc are barely distinguishable from each other, save for what language they're used in

build.sbt
LukaszKontowski: Ok, if I would be able to get rid of all warnings, then I will add this flag

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
LukaszKontowski: Issue created: https://github.com/VirtusLab/ide-probe/issues/242 - until the issue is done, I suggest to have this `@nowarn` annotation added.

build.sbt
LukaszKontowski: For now, I cannot add this flag, due to https://github.com/VirtusLab/ide-probe/issues/243

core/driver/test-plugin/src/main/scala/org/virtuslab/ideprobe/test/OpenWindowAction.scala
```diff
@@ -8,6 +8,7 @@ import javax.swing.Icon
 
 class OpenWindowAction extends AnAction("Open Window") with DumbAware {
   override def actionPerformed(e: AnActionEvent): Unit = {
+    val p = e.getProject
```
PawelLipski, line 11: Huh I'd be really surprised if that was the reason 😯

LukaszKontowski: Yeah, but test failures are so confusing that I want to try something. Looks like something from my changes is the reason, as these tests were successful on previous PR today https://github.com/VirtusLab/ide-probe/pull/240
(still, I have tha same failures for tests locally on **master** branch, but maybe this is some other reason for this)

core/probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/IntelliJApi.scala
LukaszKontowski: Looks like we would have to stick to the `isAccessible` for now. Proper replacement with `canAccess` is not that obvious. Issue https://github.com/VirtusLab/ide-probe/issues/244 created.

LukaszKontowski:
Issues created as a result of working on this PR:
https://github.com/VirtusLab/ide-probe/issues/242
https://github.com/VirtusLab/ide-probe/issues/243
https://github.com/VirtusLab/ide-probe/issues/244

PawelLipski (approved):

core/probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/Builds.scala
```diff
@@ -109,7 +109,7 @@ object Builds extends IntelliJApi {
         compileContext: CompileContext
     ): Unit = {
       def messages(category: CompilerMessageCategory): Seq[BuildMessage] = {
-        compileContext.getMessages(category).map { msg =>
+        compileContext.getMessages(category).toIndexedSeq.map { msg =>
```
lukaszwawrzyk, line 112: why this toIndexedSeq conversion happens everywhere?

LukaszKontowski: Without it (as we have on master now) - we are having following warnings (which contain answer, **why**):
```
[warn] /Users/lkontowski/code/ide-probe/core/probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/Builds.scala:112:50: method copyArrayToImmutableIndexedSeq in class LowPriorityImplicits2 is deprecated (since 2.13.0): Implicit conversions from Array to immutable.IndexedSeq are implemented by copying; Use the more efficient non-copying ArraySeq.unsafeWrapArray or an explicit toIndexedSeq call
[warn]         compileContext.getMessages(category).map { msg =>
[warn]                                                  ^
[warn] one warning found
```

PR #245, opened 2022-08-10
WIP: replaced isAccessible with canAccess

LukaszKontowski:


core/probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/IntelliJApi.scala
```diff
@@ -87,8 +87,14 @@ trait IntelliJApi {
     }
 
     private def using[B <: AccessibleObject, C](accessible: B)(f: B => C): C = {
+      val modifiers = accessible.asInstanceOf[Member].getModifiers
+      val objArgument =
+        if (!Modifier.isStatic(modifiers) && (accessible.isInstanceOf[Method] || accessible.isInstanceOf[Field]))
+          obj
+        else
+          null
```
PawelLipski, line 95: Okay... so the cases of static methods/fields are handled there... and what else?

LukaszKontowski: I'm not sure I understand, but takle a look at the snippet from the `canAccess` method. It looks like the main thing we have to do before using it is to decide whether the `obj` parameter should be a real object or null.

lukaszwawrzyk (approved):

PawelLipski (approved):

Issue #262, opened 2022-08-22
Add proper descriptions and examples for all values in `reference.conf`


LukaszKontowski:
As https://github.com/VirtusLab/ide-probe/pull/252 is already quite big - let's do this in another issue & PR. For example - there are no examples and guides for the `probe.intellij.plugins` parameter, and for a few more. All possible configurations should be described and provided as examples in comments.

LukaszKontowski:
Can be done together with https://github.com/VirtusLab/ide-probe/issues/254 - these are quite similar

PR #252, opened 2022-08-15
Replace default values in code with .conf file

LukaszKontowski:
Closes #217

PR #263, opened 2022-08-23
exclude errors by parts of stack trace

### LINKED ISSUE #228: Add ability to exclude errors by parts of stack trace (and not just error message)
To handle cases like https://app.circleci.com/pipelines/github/VirtusLab/git-machete-intellij-plugin/5603/workflows/01bccb5e-a0f7-4c22-9682-56ccc260bd6f/jobs/5910/tests : just a spurious NPE, coming somewhere from IntelliJ. We can't just filter out all `java.lang.NullPointerException`s... but filtering out by pieces of stack trace would make more sense.

LukaszKontowski:
Closes #228

Adds ability to include/exclude errors based on both one-line and multiline strings from .conf files. 

https://www.scala-lang.org/api/2.13.6/scala/util/matching/Regex.html#findFirstIn(source:CharSequence):Option[String] did not work for multiline strings, hence usage of regex replaced with `contains` call.


PawelLipski:
_ did not work for multiline strings_

Yeah, generally most regex libs/tools that I've seen across the _software universe_ have this limitation ://

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
```diff
@@ -1,13 +1,15 @@
...
+
 import org.virtuslab.ideprobe.config.CheckConfig
 import org.virtuslab.ideprobe.protocol.IdeMessage
 
 object ErrorValidator {
   def apply(config: CheckConfig, errors: Seq[IdeMessage]): Option[Exception] = {
     val filteredErrors = errors
-      .filter(error => config.errors.includeMessages.exists(_.r.findFirstIn(error.content).nonEmpty))
-      .filterNot(error => config.errors.excludeMessages.exists(_.r.findFirstIn(error.content).nonEmpty))
+      .filter(ideMessageMatchesMessagesFromConfig(_, config.errors.includeMessages))
```
PawelLipski, line 11: Pls name the `_` param; I generally recommend using `_` placeholder only when its usage is instantly obvious (like in the previous `exists`)

PawelLipski: Or alternatively, curry up `ideMessageMatchesMessagesFromConfig` and then use as
```suggestion
      .filter(ideMessageMatchesMessagesFromConfig(config.errors.includeMessages))
```

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
```diff
@@ -18,8 +20,13 @@ object ErrorValidator {
     }
   }
 
+  private def ideMessageMatchesMessagesFromConfig(ideMessage: IdeMessage, messagesFromConfig: Seq[String]): Boolean =
+    messagesFromConfig.exists(configMessage => trimMultiLine(ideMessage.content).contains(trimMultiLine(configMessage)))
+
+  private def trimMultiLine(s: String): String = s.split('\n').map(_.trim).mkString("\n")
```
PawelLipski, line 26: I'd call it
```suggestion
  private def trimEachLine(s: String): String = s.split('\n').map(_.trim).mkString("\n")
```
As `trimMultiLine` might suggest the the **entire multiline string** is trimmed, rather than each individual line

PawelLipski: Wow btw apostrophe works in Scala? 😯 I thought it was only confined to [Symbols](https://www.scala-lang.org/api/2.13.6/scala/Symbol.html) and generally deprecated

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
```diff
@@ -1,13 +1,15 @@
 package org.virtuslab.ideprobe.reporting
 
+import scala.collection.mutable
+
 import org.virtuslab.ideprobe.config.CheckConfig
 import org.virtuslab.ideprobe.protocol.IdeMessage
 
 object ErrorValidator {
```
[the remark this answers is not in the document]
LukaszKontowski: Sure, I'm planning to do that within https://github.com/VirtusLab/ide-probe/issues/262 - but first https://github.com/VirtusLab/ide-probe/pull/252 must be merged

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
LukaszKontowski: Hm, here we have just a plain `Char` usage - `s.split('\n')` ... I don't get it?

PawelLipski: Aww crap, sorry, I get JVM languages confused ;p sure, all good

PawelLipski (approved):

PawelLipski:
Btw if you could open a PR for https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1063 while you're at this

LukaszKontowski:
> Btw if you could open a PR for [VirtusLab/git-machete-intellij-plugin#1063](https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1063) while you're at this

Sure, but maybe after merging this PR? Or we want to start with some SNAPSHOT `ide-probe` draft PR?

PawelLipski:
Snapshot FTW

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
```diff
@@ -18,8 +20,13 @@ object ErrorValidator {
     }
   }
 
+  private def ideMessageMatchesMessagesFromConfig(ideMessage: IdeMessage, messagesFromConfig: Seq[String]): Boolean =
+    messagesFromConfig.exists(configMessage => trimEachLine(ideMessage.content).contains(trimEachLine(configMessage)))
+
+  private def trimEachLine(s: String): String = s.split('\n').map(_.trim).mkString("\n")
+
   private def toString(errors: Seq[IdeMessage]): String = {
-    val sb = new StringBuilder()
```
PawelLipski, line 22: Hmmm was it a different `StringBuilder` class beforehand? before the explicit import has been added

LukaszKontowski: No, it has been the same. Added explicit import as IDE suggested. IMO it's nice to get rid of IDE alerts if possible

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
```diff
@@ -1,13 +1,15 @@
...
+
 import org.virtuslab.ideprobe.config.CheckConfig
 import org.virtuslab.ideprobe.protocol.IdeMessage
 
 object ErrorValidator {
   def apply(config: CheckConfig, errors: Seq[IdeMessage]): Option[Exception] = {
     val filteredErrors = errors
-      .filter(error => config.errors.includeMessages.exists(_.r.findFirstIn(error.content).nonEmpty))
-      .filterNot(error => config.errors.excludeMessages.exists(_.r.findFirstIn(error.content).nonEmpty))
+      .filter(error => ideMessageMatchesMessagesFromConfig(error, config.errors.includeMessages))
```
PawelLipski, line 11: Hmmm for clarity... what's the default value for `includeMessages`? if the user doesn't modify the config, will **all** error messages be included by default? 🤔

LukaszKontowski:
```
includeMessages = [".*"]
excludeMessages = []
```
...
Looks like some update will be needed - ".*" was for regex matching. Thanks for spotting it. I'll add a commit soon.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
```diff
@@ -18,8 +20,16 @@ object ErrorValidator {
     }
   }
 
+  private def ideMessageMatchesMessagesFromConfig(ideMessage: IdeMessage, messagesFromConfig: Seq[String]): Boolean =
+    messagesFromConfig.exists(configMessage =>
```
PawelLipski, line 24: Nit: change to braces for better looks
```suggestion
    messagesFromConfig.exists { configMessage =>
```

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
```diff
@@ -18,8 +20,16 @@ object ErrorValidator {
...
+      trimEachLine(ideMessage.content).contains(trimEachLine(configMessage)) ||
+        configMessage.r.findFirstIn(ideMessage.content).nonEmpty
```
PawelLipski, line 26: Okay... so the match is either based on regex matching a single line, or non-regex literal string matching multiple lines? 🤔 

TBH it's getting pretty tricky/convoluted as for config key... I'd consider `exclude/includeMessages` as they are (regex, single-line) and adding a new pair of keys for matching stack trace specifically (non-regex, multi-line)

LukaszKontowski: > Okay... so the match is either based on regex matching a single line, or non-regex literal string matching multiple lines?

Well, not exactly. It will work as described below, both for `probe.driver.check.errors.includeMessages` and `probe.driver.check.errors.excludeMessages`:

1) If user adds a **regex-string** in the config file - it will work well and match error message **as long as the regex-string is NOT multiline**
2) If user adds a **string literal** (without regex usage) - it will always work both for single-line and multiline **string literal**

It might look complicated in the code, but from the user's perspective - if we add a short note in the `reference.conf` - that usage of regular expressions is limited to single-line strings - it should be easy to understand

. I would suggest to keep two config keys: `probe.driver.check.errors.includeMessages` and `probe.driver.check.errors.excludeMessages` and not add new ones. In my opinion adding more keys is the thing that would make configuration more complicated.

LukaszKontowski: And such descriptions will be added to `reference.conf` soon - in the https://github.com/VirtusLab/ide-probe/issues/262 issue

PawelLipski: Okay! makes sense indeed now. I can see some exotic edge cases with false-positives (like, the given string doesn't match when interpreted verbatim as intended, but accidentally matches as a regex)... but I wouldn't care too much

PawelLipski: Esp. that regex-special characters like `+*?{}` aren't likely to occur in stack traces; only `.` and `()` are prevalent

PawelLipski: Although.... one thing! what if `configMessage` is an **invalid regex**, like `lol (xd`? won't it end up in an exception on `.r` call?

PawelLipski: Pls add such a case to the tests, even if it doesn't end up in an exception in the current implementation

LukaszKontowski: tested with a `"at com.intellij.psi.impl.file.impl.FileManagerImpl.getCachedPsiFile(F"` string...
```
scala> "at com.intellij.psi.impl.file.impl.FileManagerImpl.getCachedPsiFile(F".r
java.util.regex.PatternSyntaxException: Unclosed group near index 69
at com.intellij.psi.impl.file.impl.FileManagerImpl.getCachedPsiFile(F
  at java.base/java.util.regex.Pattern.error(Pattern.java:2028)
  at java.base/java.util.regex.Pattern.accept(Pattern.java:1878)
  at java.base/java.util.regex.Pattern.group0(Pattern.java:3053)
  at java.base/java.util.regex.Pattern.sequence(Pattern.java:2124)
  at java.base/java.util.regex.Pattern.expr(Pattern.java:2069)
  at java.base/java.util.regex.Pattern.compile(Pattern.java:1783)
  ... 63 elided
```
It will throw an exception. But maybe such exception would be a good thing - so that user will know that the configuration is invalid? `ide-probe` is used for tests so throwing such Exception might not be a bad thing? (it would be bad if `ide-probe` was for runtime...)

LukaszKontowski: And yet, before my changes - we had the same possibility for exceptions here - we were just calling `.r` on each string from these configs

PawelLipski: Right! makes sense indeed, let's leave as is. Up to you whether a test case is needed (maybe not)

LukaszKontowski: On the other hand, such scenario seems impossible to me. Since || short circuits and 
`trimEachLine(ideMessage.content).contains(trimEachLine(configMessage))`
would evaluate to `true`... Exception would not be thrown.

PawelLipski: I mean, in case where the first operand `||` evaluates to false, it's still possible that `.r` is called, right?

PawelLipski (approved):

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
```diff
@@ -18,8 +20,16 @@ object ErrorValidator {
     }
   }
 
+  private def ideMessageMatchesMessagesFromConfig(ideMessage: IdeMessage, messagesFromConfig: Seq[String]): Boolean =
+    messagesFromConfig.exists { configMessage =>
+      trimEachLine(ideMessage.content).contains(trimEachLine(configMessage)) ||
+      configMessage.r.findFirstIn(ideMessage.content).nonEmpty
+    }
+
+  private def trimEachLine(s: String): String = s.split('\n').map(_.trim).mkString("\n")
```
lukaszwawrzyk, line 29: there is `linesIterator` method

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/ErrorValidatorTest.scala
```diff
@@ -0,0 +1,119 @@
...
+import org.virtuslab.ideprobe.config.CheckConfig.ErrorConfig
+import org.virtuslab.ideprobe.protocol.IdeMessage
+import org.virtuslab.ideprobe.protocol.IdeMessage.Level
+import org.virtuslab.ideprobe.reporting.ErrorValidator
+
+class ErrorValidatorTest {
+  private val configRoot = "probe"
+  private val defaultConfigWithErrorsEnabled: Config = Config.fromClasspath("error_validator.conf")
+  private val defaultCheckConfigWithErrorsEnabled: CheckConfig =
+    IntelliJFixture.readIdeProbeConfig(defaultConfigWithErrorsEnabled, configRoot).driver.check
```
lukaszwawrzyk, line 20: Maybe just read the config from string in test. It will be easier to follow.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/ErrorValidatorTest.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  private val messageBusConnectionImplNPE = fromResource("ideError_1.txt").mkString
+  private val uastMetaLanguageNPE = fromResource("ideError_2.txt").mkString
+
+  private val errorIdeMessages: List[IdeMessage] = List(messageBusConnectionImplNPE, uastMetaLanguageNPE)
+    .map(content => IdeMessage(level = Level.Error, content, None))
```
lukaszwawrzyk, line 26: you don't really need named parameter for `level` but it would be useful for that `None`.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/ErrorValidatorTest.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  private val messageBusSpecificString = "at com.intellij.util.messages.impl.MessageBusConnectionImpl"
+  private val uastMetaSpecificString = "at com.intellij.uast.UastMetaLanguage"
+
+  @Test
+  def shouldIncludeAllErrorsByDefault(): Unit = {
+    val exceptions: List[Option[Exception]] = collectExceptions(defaultCheckConfigWithErrorsEnabled)
+    val exceptionsIncluded = exceptions.filter(_.nonEmpty)
```
lukaszwawrzyk, line 34: you can use `partition` method

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
LukaszKontowski: sure, and what would be the benefit from using it instead of current logic?

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/ErrorValidatorTest.scala
```diff
@@ -0,0 +1,119 @@
...
+    val exceptionsIgnored = exceptions.filter(_.isEmpty)
+    assertEquals(2, exceptionsIncluded.size)
+    assertEquals(0, exceptionsIgnored.size)
+  }
+
+  @Test
+  def shouldIgnoreAllErrorsIfNotEnabled(): Unit = {
+    val checkConfigErrorsDisabled = defaultCheckConfigWithErrorsEnabled.copy(errors =
+      ErrorConfig(
```
lukaszwawrzyk, line 43: `defaultCheckConfigWithErrorsEnabled.errors.copy(enabled = false)` maybe

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/ErrorValidatorTest.scala
```diff
@@ -0,0 +1,119 @@
...
+        enabled = defaultCheckConfigWithErrorsEnabled.errors.enabled,
+        includeMessages = defaultCheckConfigWithErrorsEnabled.errors.includeMessages,
+        excludeMessages = Seq(s".*$messageBusSpecificString.*")
+      )
+    )
+    val exceptions: List[Option[Exception]] = collectExceptions(checkConfigToIgnoreMessageBusErrors)
+    val exceptionsIncluded = exceptions.filter(_.nonEmpty)
+    val exceptionsIgnored = exceptions.filter(_.isEmpty)
+    assertEquals(1, exceptionsIncluded.size)
```
lukaszwawrzyk, line 85: I guess assertion logic in these tests could be less complicated. Just return `List[Exception]` and check if interesting ones are there and if count is right. Now each test is validating that if you pass a list of 2 elements to a helper method that you added in this test, it returns 2 results. Ofc it will, it just calls map and it is not a part of tested code :p

LukaszKontowski: > Now each test is validating that if you pass a list of 2 elements to a helper method that you added in this test, it returns 2 results. Ofc it will, it just calls map and it is not a part of tested code :p

Hm, I don't think your're right here. In each test `exceptionsIncluded.size` and `exceptionsIgnored .size` are checked. To make sure, that including / ignoring errors works fine. For example:
- `shouldIncludeAllErrorsByDefault()` - tests as follows:
```
assertEquals(2, exceptionsIncluded.size)
assertEquals(0, exceptionsIgnored.size)
```
- `shouldExcludeSpecificErrorsByRegexUsage()` - tests as follow:
```
assertEquals(1, exceptionsIncluded.size)
assertEquals(1, exceptionsIgnored.size)
```
and it is a part of tested code - we make sure that ErrorValidator includes / excludes errors properly with the config.

Or maybe I just didn't understand your comment properly?

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/ErrorValidatorTest.scala
LukaszKontowski: well... Why make it more complicated by nesting another `.copy` call? IMO just `enabled = false` is much more readible

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
lukaszwawrzyk: well, it was created for the purpose of splitting string into lines. It is lazy and likely covers windows/linux separators. Also it names clearly shows what you mean. Even if you don't need these properties, why reimplement it if it is there?

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/ErrorValidatorTest.scala
lukaszwawrzyk: You could have lenses to have it more concise. As they are not here yet, nested copying is unavoidable. What you did here is inlining of copy method. Why reimplementing a method that already exists. If we add more fields to ErrorConfig, this place would need to be changed. Also you clearly state that you are changing just the `enabled` property, while otherwise reader needs to think what are you putting in this `included` and `excludedMessages` (is it just the same, or does it change as well?) and this code doesn't need to be there.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/ErrorValidatorTest.scala
lukaszwawrzyk: I mean, it is enough to check how many errors were found. Otherwise you are testing the test code. It is clear that `ignored count == 2 - error count` if you pass w errors. You don't need to complicate test code with optionals and get calls.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/reporting/ErrorValidator.scala
LukaszKontowski: ok, done as suggested

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/ErrorValidatorTest.scala
LukaszKontowski: ok, done as suggested

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/ErrorValidatorTest.scala
LukaszKontowski: ok, done as suggested

lukaszwawrzyk (approved):

PR #266, opened 2022-08-26
Make idea logs accessible after shutdown

### LINKED ISSUE #226: Instance directory does not exist once instance is shutdown
During ui tests execution within a running instance, there is a instance directory (`intellij-instance-222.3345.16-EAP-SNAPSHOT--FFkv-SzpR2mJxAtNCAHFdQ`) and a log file available (`idea.log`).
```
/root/.ideprobe-uitests/instances/intellij-instance-222.3345.16-EAP-SNAPSHOT--FFkv-SzpR2mJxAtNCAHFdQ/logs/idea.log
```
---
UI test run finishes (gradle task), instance shuts down, and as a result instance directory is empty.
```
/root/.ideprobe-uitests/instances
```

I need to access `idea.log` file. I think that the instance dir (and it's content) should be kept. 

(I found it while working on https://github.com/VirtusLab/git-machete-intellij-plugin/pull/984)

LukaszKontowski:
Closes #226

### DIFF
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -126,12 +126,11 @@ probe {
     // By default, `trusted` is resolved as the root directory: "/".
     # trusted = "/"
     //
-    // `diagnostics` is a directory where diagnostics data, like logs from the intelliJ instance, can be copied into
-    // before cleanup of the IntelliJ instance. Specify proper path if you want to have access to these
-    // files after tests finish (could be useful when working with CI pipelines). By default, `diagnostics` directory
-    // is NOT created and such data is deleted from the disk with all other files from the IntelliJ instance directory.
-    // If you use this parameter, the `diagnostics` directory will be created and it will NOT be removed/cleared.
-    // Example path below:
-    # diagnostics = "/tmp/ide-probe/diagnostics-data"
+    // `logExport` is a directory where logs from the intelliJ instance can be copied into before cleanup
+    // of the IntelliJ instance. Specify proper path if you want to have access to IDEA logs after tests finish.
+    // By default, `logExport` directory is NOT created and such data is deleted from the disk with all other files
+    // from the IntelliJ instance directory. If you use this parameter, the `logExport` directory will be created
+    // and it will NOT be removed/cleared. Example path below:
+    # logExport = "/tmp/ide-probe/idea-logs"
   }
 }
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -125,5 +125,13 @@ probe {
     // treat all projects/repositories under the `trusted` directory as "safe" and will run the code without alerts.
     // By default, `trusted` is resolved as the root directory: "/".
     # trusted = "/"
+    //
+    // `diagnostics` is a directory where diagnostics data, like logs and cache from the intelliJ instance, can
+    // be copied into before cleanup of the IntelliJ instance. Specify proper path if you want to have access to these
+    // files after tests finish (could be useful when working with CI pipelines). By default, `diagnosticsData` directory
+    // is NOT created and such data is deleted from the disk with all other files from the IntelliJ instance directory.
+    // If you use this parameter, the `diagnostics` directory will be created and it will NOT be removed/cleared.
+    // Example path below:
+    # diagnostics "/tmp/ide-probe/diagnostics-data"
   }
 }
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -126,12 +126,12 @@ probe {
     // By default, `trusted` is resolved as the root directory: "/".
     # trusted = "/"
     //
-    // `diagnostics` is a directory where diagnostics data, like logs and cache from the intelliJ instance, can
-    // be copied into before cleanup of the IntelliJ instance. Specify proper path if you want to have access to these
-    // files after tests finish (could be useful when working with CI pipelines). By default, `diagnosticsData` directory
+    // `diagnostics` is a directory where diagnostics data, like logs from the intelliJ instance, can be copied into
+    // before cleanup of the IntelliJ instance. Specify proper path if you want to have access to these
+    // files after tests finish (could be useful when working with CI pipelines). By default, `diagnostics` directory
     // is NOT created and such data is deleted from the disk with all other files from the IntelliJ instance directory.
     // If you use this parameter, the `diagnostics` directory will be created and it will NOT be removed/cleared.
     // Example path below:
-    # diagnostics "/tmp/ide-probe/diagnostics-data"
+    # diagnostics = "/tmp/ide-probe/diagnostics-data"
   }
 }
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IdeProbePaths.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IdeProbePaths.scala
@@ -11,13 +11,14 @@ case class IdeProbePaths(
     workspaces: Path,
     screenshots: Path,
     cache: Path,
-    trusted: Path
+    trusted: Path,
+    diagnostics: Option[Path]
 )
 
 object IdeProbePaths {
   val Default: IdeProbePaths = {
     // TODO: replace None parameters with loading defaults from reference.conf
-    from(PathsConfig(None, None, None, None, None, None))
+    from(PathsConfig(None, None, None, None, None, None, None))
   }
 
   def from(config: PathsConfig): IdeProbePaths = {
@@ -27,7 +28,8 @@ object IdeProbePaths {
     val screenshotsPath = config.screenshots.getOrElse(basePath.resolve("screenshots"))
     val cachePath = config.cache.getOrElse(basePath.resolve("cache"))
     val trustedPath = config.trusted.getOrElse(Paths.get("/"))
+    val diagnosticsOption = config.diagnostics
 
-    IdeProbePaths(basePath, instancesPath, workspacesPath, screenshotsPath, cachePath, trustedPath)
+    IdeProbePaths(basePath, instancesPath, workspacesPath, screenshotsPath, cachePath, trustedPath, diagnosticsOption)
   }
 }
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IdeProbePaths.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IdeProbePaths.scala
@@ -28,8 +28,8 @@ object IdeProbePaths {
     val screenshotsPath = config.screenshots.getOrElse(basePath.resolve("screenshots"))
     val cachePath = config.cache.getOrElse(basePath.resolve("cache"))
     val trustedPath = config.trusted.getOrElse(Paths.get("/"))
-    val diagnosticsOption = config.diagnostics
+    val diagnosticsPath = config.diagnostics
 
-    IdeProbePaths(basePath, instancesPath, workspacesPath, screenshotsPath, cachePath, trustedPath, diagnosticsOption)
+    IdeProbePaths(basePath, instancesPath, workspacesPath, screenshotsPath, cachePath, trustedPath, diagnosticsPath)
   }
 }
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IdeProbePaths.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IdeProbePaths.scala
@@ -12,7 +12,7 @@ case class IdeProbePaths(
     screenshots: Path,
     cache: Path,
     trusted: Path,
-    diagnostics: Option[Path]
+    logExport: Option[Path]
 )
 
 object IdeProbePaths {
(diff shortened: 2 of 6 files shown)


core/driver/sources/src/main/resources/reference.conf
```diff
@@ -125,5 +125,13 @@ probe {
...
     // By default, `trusted` is resolved as the root directory: "/".
     # trusted = "/"
+    //
+    // `diagnostics` is a directory where diagnostics data, like logs and cache from the intelliJ instance, can
+    // be copied into before cleanup of the IntelliJ instance. Specify proper path if you want to have access to these
+    // files after tests finish (could be useful when working with CI pipelines). By default, `diagnosticsData` directory
+    // is NOT created and such data is deleted from the disk with all other files from the IntelliJ instance directory.
+    // If you use this parameter, the `diagnostics` directory will be created and it will NOT be removed/cleared.
+    // Example path below:
+    # diagnostics "/tmp/ide-probe/diagnostics-data"
```
PawelLipski, line 135:
```suggestion
    # diagnostics = "/tmp/ide-probe/diagnostics-data"
```

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -125,5 +125,13 @@ probe {
     // treat all projects/repositories under the `trusted` directory as "safe" and will run the code without alerts.
     // By default, `trusted` is resolved as the root directory: "/".
     # trusted = "/"
+    //
+    // `diagnostics` is a directory where diagnostics data, like logs and cache from the intelliJ instance, can
+    // be copied into before cleanup of the IntelliJ instance. Specify proper path if you want to have access to these
+    // files after tests finish (could be useful when working with CI pipelines). By default, `diagnosticsData` directory
```
PawelLipski, line 131:
```suggestion
    // files after tests finish (could be useful when working with CI pipelines). By default, `diagnostics` directory
```

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -219,5 +219,20 @@ final class DownloadedIntelliJ(
...
+    probePaths.diagnostics match {
+      case Some(path) =>
+        root
+          .resolve("logs")
+          .copyDir(path.resolve(getPathWithVersionNumber(root).getFileName).resolve("logs"))
+        root.delete()
+      case None =>
+        root.delete()
+    }
+    root.delete()
```
PawelLipski, line 232: Hmmm double delete root?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -219,5 +219,20 @@ final class DownloadedIntelliJ(
...
+  }
+
+  private def getPathWithVersionNumber(intellijRootPath: Path): Path =
+    if (intellijRootPath.getFileName.toString == "Contents") intellijRootPath.getParent else intellijRootPath
```
PawelLipski, line 236: Hmmm what special case is handled here? 🤔

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
LukaszKontowski: yeah, mistake - will fix it in next commit

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/SingleRunFixtureTest.scala
```diff
@@ -87,4 +89,17 @@ final class SingleRunFixtureTest extends IdeProbeFixture with WorkspaceFixture w
...
+  def copiesDiagnosticsDataIntoConfiguredDirectoryBeforeCleanup(): Unit = {
+    val config = Config.fromString("""probe.paths.diagnostics = "/tmp/ide-probe-test"""")
+    val fixtureFromConfig = IntelliJFixture.fromConfig(config)
+    val diagnosticsFixture = new SingleRunIntelliJ(fixtureFromConfig)
+
+    diagnosticsFixture { runningIntellijFixture =>
+      // nothing special to be done here
+    }
+
+    assertTrue(Paths.get(s"/tmp/ide-probe-test").directChildren().nonEmpty)
```
PawelLipski, line 103: Maybe check for `logs` child dir explicitly?

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -125,5 +125,13 @@ probe {
     // treat all projects/repositories under the `trusted` directory as "safe" and will run the code without alerts.
     // By default, `trusted` is resolved as the root directory: "/".
     # trusted = "/"
+    //
+    // `diagnostics` is a directory where diagnostics data, like logs and cache from the intelliJ instance, can
```
PawelLipski, line 129: Is cache also going to be included/is it going to be useful?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
PawelLipski: Also, it's not clear why it's a path **with version number**

LukaszKontowski: If I get it right, there is a difference:
- for Linux based OS `root` will be something like `intellij-instance-2022.2.1--T3ySdgShSvyr87HNoJq-oQ`
- for macOs `root` will be something like `intellij-instance-2022.2.1--T3ySdgShSvyr87HNoJq-oQ/Contents`

PawelLipski: Wow okay 🍎 pls document in code comment

PawelLipski: Also, isn't the _version number_ just an internal implementation detail? as in, do we care whether the path contains the version or not, as long as it point to a valid IDE dir?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IdeProbePaths.scala
```diff
@@ -27,7 +28,8 @@ object IdeProbePaths {
     val screenshotsPath = config.screenshots.getOrElse(basePath.resolve("screenshots"))
     val cachePath = config.cache.getOrElse(basePath.resolve("cache"))
     val trustedPath = config.trusted.getOrElse(Paths.get("/"))
+    val diagnosticsOption = config.diagnostics
```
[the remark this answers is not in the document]
LukaszKontowski: No, let me change to stick to the convention.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
LukaszKontowski: Hm, if there would be multiple tests running in one process on the same instance (machine / docker instance etc.) and these different tests would use different versions of intellij - it would be nice to have logs grouped by intellij versions at least. That's why I decided o take it into account. Does it make sense? Or we need some other logic for such scenarios?

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: Hm, what cache directory? I did not find such directory inside the IntelliJ instance dir:
```
lkontowski@lkontowski-MacBook-Pro /Users/lkontowski 
% ls -1 /Applications/IntelliJ\ IDEA\ CE.app/Contents
CodeResources
Info.plist
MacOS
Resources
_CodeSignature
bin
jbr
lib
license
plugins
```

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/SingleRunFixtureTest.scala
LukaszKontowski: Changed tests logic - please take a look

core/driver/sources/src/main/resources/reference.conf
PawelLipski: I mean... it's mentioned in your comment ;)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
PawelLipski: Ok, makes sense

core/driver/sources/src/main/resources/reference.conf
PawelLipski: Oh okay, removed

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -125,5 +125,13 @@ probe {
...
     // By default, `trusted` is resolved as the root directory: "/".
     # trusted = "/"
+    //
+    // `diagnostics` is a directory where diagnostics data, like logs from the intelliJ instance, can be copied into
+    // before cleanup of the IntelliJ instance. Specify proper path if you want to have access to these
+    // files after tests finish (could be useful when working with CI pipelines). By default, `diagnostics` directory
+    // is NOT created and such data is deleted from the disk with all other files from the IntelliJ instance directory.
+    // If you use this parameter, the `diagnostics` directory will be created and it will NOT be removed/cleared.
+    // Example path below:
+    # diagnostics = "/tmp/ide-probe/diagnostics-data"
```
PawelLipski, line 135: I'm uncertain about this `diagnostics` name... how about `exportLogsTo` or `logExport`? WDYT Łukaszes?

PawelLipski: I'd just like to differentiate the paths passed to IntelliJ as configs (all others) from this one, which isn't an IntelliJ config path per se.

LukaszKontowski: Good idea! Renamed to logExport.

PawelLipski (approved):

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -219,5 +219,26 @@ final class DownloadedIntelliJ(
   override val ideaProperties: Path =
     root.resolve("bin").resolve("idea.properties").write(ideaPropertiesContent)
 
-  override def cleanup(): Unit = root.delete()
+  override def cleanup(): Unit =
+    probePaths.logExport match {
+      case Some(path) =>
+        root
```
lukaszwawrzyk, line 225: use `paths.logs`

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/SingleRunFixtureTest.scala
```diff
@@ -87,4 +89,26 @@ final class SingleRunFixtureTest extends IdeProbeFixture with WorkspaceFixture w
...
+    val logExportFixture = new SingleRunIntelliJ(fixtureFromConfig)
+
+    logExportFixture { _ =>
+      // nothing special to be done here, invoked just for cleanup
+    }
+
+    val filesAndDirsFromLogExportDirectory = Paths.get(tmpDirString).recursiveChildren()
+
+    assertTrue(filesAndDirsFromLogExportDirectory.exists { path =>
+      path.getFileName.toString == "logs" && path.isDirectory && path.directChildren().nonEmpty
```
lukaszwawrzyk, line 107: just use path.name from PathExtensions

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -219,5 +219,26 @@ final class DownloadedIntelliJ(
...
+          .resolve("logs")
+          .copyDir(path.resolve(getPathWithVersionNumber(root).getFileName).resolve("logs"))
```
lukaszwawrzyk, line 227: just make this method return the name. Also use .name from Path extensions

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -219,5 +219,26 @@ final class DownloadedIntelliJ(
...
+        root.delete()
+      case None =>
+        root.delete()
```
lukaszwawrzyk, line 230: if you always call root.delete() you can move it outside this pattern matching to not repeat it. probePaths.logExport.foreach  { path =>
  // do the copy
}

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
LukaszKontowski: But in this case I need to use the `copyDir` method, which needs a **Path** not a **String**. Moreover, the `.resolve("logs")` part also must be invoked on an object of type **Path**. That's why the `getPathWithVersionNumber` method should return a **Path**.

lukaszwawrzyk: You are calling `resolve` on `path`. Resolve accepts both `Path` and `String` so nothing should stop you from changing `getPathWithVersionNumber(root).getFileName` to for example: `getIntelliJInstanceIdentifier(root)` - where the new method would look just like old one but with `.name` call before return.

LukaszKontowski: ok, done 👍

lukaszwawrzyk (approved):

PR #267, opened 2022-08-26
Fixing problem with creating run configuration twice

sellophane:


core/probePlugin/src/main/scala/org/virtuslab/ideprobe/handlers/RunConfigurations.scala
```diff
@@ -84,6 +86,12 @@ object RunConfigurations extends IntelliJApi {
     }
     val selectedConfiguration = producer.getConfigurationSettings
 
+    val configurationContext = ConfigurationContext.getFromContext(dataContext)
```
lukaszwawrzyk, line 89: I think this line can be deleted, it was creating this single unnecessary config. Then `createDataContext` can go back to be part of `availableRunConfigurations`. There must be other ways to get RunManager, either from its static method or maybe from `producer` or `selectedConfiguration`.

lukaszwawrzyk (changes requested):

lukaszwawrzyk (approved):

Issue #253, opened 2022-08-17
Replace default values of `IntelliJFixture` with loading from `reference.conf`


LukaszKontowski:

PawelLipski:
In fact, I'd do this particular thing within #252 as well... IIUC this is a subset of what #252 is about 🤔

LukaszKontowski:
Ok, let's keep PRs small

PawelLipski:
Esp. for `probe.resolver.intellij.{jbr,repositories}`... they shouldn't be empty lists in reference.conf

PR #179, opened 2022-05-11
Add support for unzipped IntelliJs in resolvers.intellij.repositories config

wiacekm:
Automatically recognise if it is a zip archive or directory.

Closes: #158

PR #268, opened 2022-09-01
Draft: Replace IntelliJProvider.Default usages with values from config

LukaszKontowski:
Closes https://github.com/VirtusLab/ide-probe/issues/253
- removed default values from scala code, if these values could be read from `reference.conf`
- added one new **optional** key to the config: `probe.intellij.version.format` - so that `core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala` would be able to resolve the pattern to a `.zip` or `.dmg` file - depending on parsed configuration
-  removed `core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJDmgResolver.scala` file - as now users can use `.dmg` distributions by using mentioned `probe.intellij.version.format` config (no need to keep duplicated logic - let's try to make users use `.conf` file instead of direct calls to parts of our scala code where possible)
- removed most of logic from `core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJZipResolver.scala` along with the file and moved the necessary `def fromConfig(config: DependenciesConfig.IntelliJ): Seq[DependencyResolver[IntelliJVersion]]` to the `IntelliJResolver` companion object.
- changed default `probe.driver.launch.command` from empty Array ( `[]` ) to a one-element array: `["idea"]` - which contains the basic command that is used anyway (changed to `idea.sh` in our scala code if needed). Such change makes it easier to understand for users, that under this config they can provide their own command that starts the process.
- added `fromReferenceConf` method to `object Config` - to decrease boilerplate and keep it possible for users to use the "official" / "default" keywords in the `probe.resolvers.intellij.repositories` config
- removed some replacements from `IntelliJPatternResolver` - the ones which were always resolved to the same value
- updated unit tests accordingly

Changes moved to a new PR for clear commit history - https://github.com/VirtusLab/ide-probe/pull/275


core/driver/sources/src/main/resources/reference.conf
```diff
@@ -52,18 +52,27 @@ probe {
...
-    // default configuration for `plugins` is Plugins(None) - adding possible parameters in comments for reference
-    plugins {
-      // `repository.uri` is the URI of the plugin repository - use this config if you want to use a non-standard
-      // plugin repository. The standard one used by ide-probe is: https://plugins.jetbrains.com/plugin/download
-      # repository.uri = ""
-    }
-    jbr {
-      repositories = []
-    }
+    intellij.repositories = [
```
PawelLipski, line 55: Nice, so much clearer now 🥰

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -52,18 +52,27 @@ probe {
...
+      "https://download.jetbrains.com/idea/nightly/ideaIC-[revision].portable.zip",
+      "https://download.jetbrains.com/idea/ideaIC-[revision].portable.zip",
+      "https://www.jetbrains.com/intellij-repository/releases/[orgPath]/[module]/ideaIC/[revision]/ideaIC-[revision].zip",
+      "https://www.jetbrains.com/intellij-repository/snapshots/[orgPath]/[module]/ideaIC/[revision]/ideaIC-[revision].zip"
+    ]
+    // `plugins.repository.uri` is the URI of the plugin repository. Use this config if you want to use a non-standard
+    // plugin repository. The standard one is https://plugins.jetbrains.com/plugin/download
+    plugins.repository.uri = "https://plugins.jetbrains.com/plugin/download"
```
PawelLipski, line 63: Pls add some newlines to separate the keys visually (esp. needed for the places where HOCON isn't colored, like GitHub)

LukaszKontowski:
If tests will pass, I will undraft it :)

core/driver/sources/src/test/resources/example.conf
```diff
@@ -2,7 +2,7 @@ probe {
   resolvers {
     intellij {
       repositories = [
-        "https://www.jetbrains.com/intellij-repository/snapshots/com/jetbrains/intellij/idea/ideaIC/[revision]/ideaIC-[revision].zip"
+        "https://www.jetbrains.com/intellij-repository/snapshots/com/jetbrains/intellij/idea/ideaIC/[revision]/ideaIC-[revision][frormat]"
```
PawelLipski, line 5: Whoops `frormat`?

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -79,6 +79,16 @@ probe {
     //
     // If you want to use the Ultimate edition, replace "ideaIC" occurences with "ideaIU" in your config file. Community
     // edition ("ideaIC") is the default one.
+    //
+    // You can also add your custom intellij repository on top of the 4 default repositories, so that ide-probe will
```
PawelLipski, line 83: Okay, so given our discussion on Slack, this entire comment is **not** certain at this point, right?

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -79,6 +79,16 @@ probe {
...
+    // first try using the IntelliJs downloaded from your repository if present. If ide-probe doesn't find the IntelliJ
+    // distribution in your repository, it will use the default 4 repositories listed below. To achieve that, specify
+    // your custom repository pattern and add the `official` or `default` keyword on another line. Example below:
+    // intellij.repositories = [
+    //   "file:///"${HOME}"/.my_intellij_idea_downloads/[revision]/ideaIC-[revision][format]",
```
PawelLipski, line 88: If it makes sense, you can actually use `.pluginVerifier` as an example (with a link to https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#tasks-runpluginverifier)

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: Hm, no - my comment is right. But my example is different - in the example from comments we have:
```
intellij.repositories = [
  "file:///"${HOME}"/.my_intellij_idea_downloads/[revision]/ideaIC-[revision][format]",
  official
]
```
so - a pattern that points to a `.zip` or `.dmg` file with intelliJ

LukaszKontowski: And here: https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/src/uiTest/resources/ideprobe.conf#L43 - we have an usage that points to a directory with installed IntelliJ

LukaszKontowski:
Let me keep this PR a draft until I will be sure about the https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/src/uiTest/resources/ideprobe.conf#L43 usage.

core/driver/sources/src/main/resources/reference.conf
PawelLipski: Okay... AFAICS since https://github.com/VirtusLab/ide-probe/pull/179 we're able to pass unzipped folders as `intellij.resolvers.intellij.repositories` right?

LukaszKontowski: Let me check that and make sure

LukaszKontowski: Ok, you are right - we're able to pass unzipped folders as `intellij.resolvers.intellij.repositories` - example of such successful tests with SNAPSHOT ide-probe version from this PR:
https://github.com/VirtusLab/git-machete-intellij-plugin/pull/1107

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -5,23 +5,35 @@ probe {
...
     version {
+
       // `release` is the "Version" and connected `build` is the "Build Number" of IntelliJ IDEA version. Both should
       // be provided and they should match. You can search for proper string pairs here:
       // https://www.jetbrains.com/intellij-repository/releases/ under "com.jetbrains.intellij.idea" heading.
       // If you want to change the default intellij version, provide both parameters in your .conf file.
       build = "212.5080.55"
       release = "2021.2.1"
+
+      // `format` stands for the file format (file extension). You can choose, which type should be downloaded.
```
PawelLipski, line 20:
```suggestion
      // `format` stands for the file format (file extension). You can choose which type should be downloaded.
```

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -5,23 +5,35 @@ probe {
...
+      // By default, ".zip" distributions of IntelliJ IDEA are downloaded. Non-default supported format is ".dmg".
+      // Other formats are not supported. You can choose only one format.
+      format = ".zip"
+      # format = ".dmg"
     }
-    // Use below config with proper path if you want to use an existing Intellij instead of downloading a version.
+
+    // Use `path` config with proper path if you want to use an existing Intellij instead of downloading a version.
```
PawelLipski, line 27: Pls mention what's the relation b/w this config key, and `file:///...` entries in `resolvers.intellij.repositories`... which one takes precedence? Actually, `intellij.path` is somewhat redundant, given that you can specify a `file:///` path in `resolvers.intellij.repositories`, amirite?

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -34,43 +46,96 @@ probe {
...
+    //   "file:///"${HOME}"/.my_intellij_idea_downloads/[revision]/ideaIC-[revision][format]",
+    //   official
+    // ]
+    //
+    // Moreover, you can add a pattern pointing to a directory of an installed IntelliJ instance instead of
+    // pattern pointing to a .zip/.dmg file. ide-probe will recognize if the pattern points to a directory with
+    // installed IntelliJ or to a .zip/.dmg file. For example:
+    // intellij.repositories = [
+    //   "file:///"${HOME}"/.pluginVerifier/ides/IC-[revision]/",
+    //   official
```
PawelLipski, line 97: Hmmm it's not explained what `official` does 🤔

LukaszKontowski: Ok, but I added a description in comments:
```
If ide-probe doesn't find the IntelliJ distribution in your repository, it will use the default 4 repositories listed below. To achieve that, specify your custom repository pattern and add the `official` or `default` keyword on another line. Example below:
```
Is it not obvious enough?

PawelLipski: Oh! okay, didn't notice... makes sense, should be pretty self-explanatory

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: `probe.intellij.path` takes precedence. 

For example, if you would specify two different paths under `probe.intellij.path` and `probe.resolvers.intellij.repositories` - then the first one would be used to install IntelliJ. Using `probe.intellij.path` tells `ide-probe` to use an `org.virtuslab.ideprobe.ide.intellij.ExistingIntelliJ` instance instead of the `org.virtuslab.ideprobe.ide.intellij.IntelliJFactory`. This means that there will be no real usage of `probe.resolvers.intellij.repositories` value in such case. I find it rather unlikely that somebody would do such a thing in their `.conf` file, but I will add a comment about such case.

In my opinion - we can leave both configs for now and possibly clean this mess in some future issue. It would be nice to have only one config where a path to existing IntelliJ instance can be passed. Currently it is somehow confusing. 

The `probe.intellij.path` way is more obvious and could be used for simple cases, where someone would like to use **only** an existing IntelliJ instance for tests.

The `probe.resolvers.intellij.repositories` way is more complicated, but gives more possibilities - like in current git-machete-intellij-plugin solution (first - try to use an existing intelliJ and if not possible - download from the internet).

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
```diff
@@ -31,14 +33,11 @@ final case class IntelliJVersion(build: String, release: Option[String]) {
...
-  val Latest: IntelliJVersion = release("2021.2.1", "212.5080.55")
-
   def snapshot(build: String): IntelliJVersion = {
-    IntelliJVersion(build, None)
+    IntelliJVersion(build = build, release = None, format = Some(configConvert.map(_.format).toString))
   }
 
   def release(version: String, build: String): IntelliJVersion = {
-    IntelliJVersion(build, Some(version))
+    IntelliJVersion(build = build, release = Some(version), format = Some(configConvert.map(_.format).toString))
```
lukaszwawrzyk, line 41: why this strange code?

core/driver/sources/src/test/resources/example.conf
```diff
@@ -2,7 +2,7 @@ probe {
   resolvers {
     intellij {
       repositories = [
-        "https://www.jetbrains.com/intellij-repository/snapshots/[orgPath]/[module]/[artifact]/[revision]/[artifact]-[revision].zip"
+        "https://www.jetbrains.com/intellij-repository/snapshots/com/jetbrains/intellij/idea/ideaIC/[revision]/ideaIC-[revision][format]"
```
lukaszwawrzyk, line 5: the test was checking if the `orgPath` `module` and `artifact` are set correctlly, why inline those?

lukaszwawrzyk:
> * removed default values from scala code, if these values could be read from `reference.conf`

The point of this task IMo was to make IntelliJProvicer.Default initialize its value from the config, not to remove them at all.
Probe can be used both from config and from code and both should be convenient.

> * added one new **optional** key to the config: `probe.intellij.version.format` - so that `core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala` would be able to resolve the pattern to a `.zip` or `.dmg` file - depending on parsed configuration

To be closer to resolvers/patterns in jvm world, this should be named ext or extension.

> * removed `core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJDmgResolver.scala` file - as now users can use `.dmg` distributions by using mentioned `probe.intellij.version.format` config (no need to keep duplicated logic - let's try to make users use `.conf` file instead of direct calls to parts of our scala code where possible)

Not sure why. Let's allow users to choose if they prefer to use hocon or scala. I have the code that relies on that. I need to resolve dmg files in one place and zip files in another place, at the same time I am adding my own resolvers. Then I use resolvers api to fetch IntelliJ. Please don't drop this api.

```
override protected val dmgResolvers =
      Seq(OfficialIntelliJDmgResolver, OfficialNightlyIntelliJDmgResolver) ++
        localIntelliJRepository.map(new IntelliJDmgResolver(_))
```

> * removed some replacements from `IntelliJPatternResolver` - the ones which were always resolved to the same value

They are now resolved to the same value because it is IntelliJPatternResolver, but we could add support for android studio or pycharm and these patterns would still be useful.

----------

Please do not drop scala code usecase, do not drop `val Default` from components (just build them form config) and default args for IdeProbeFixture.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
PawelLipski: _strange_? 🤔

LukaszKontowski: @lukaszwawrzyk - please let me know, what exactly is strange / not OK with this change. Named parameters are not OK? Or maybe the fact, that third parameter has been added? Or something else maybe?

core/driver/sources/src/test/resources/example.conf
LukaszKontowski: I changed it as [orgPath], [module] and [artifact] were always resolved to the same constant string values. Should I revert this change? Add these replacaments back to the scala code?

PawelLipski: yeah, +1, revert (it might be sth else than `ideaIC` indeed, even just `ieadIU`)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
lukaszwawrzyk: Why do you need to use configConvert explicitly in the code.

core/driver/sources/src/test/resources/example.conf
lukaszwawrzyk: yeah, for example iu/ic is controlled by resolvers, but we might add `type/product` key to config and pick specific variant.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
LukaszKontowski: It's an easy way to get the `probe.intellij.format` value from the `.conf` file. As we already have this val in the `IntelliJVersion` object - it looks like we can just read this config from the `configConvert` val. Is there something wrong with this?

LukaszKontowski: BTW - what is the actual usage of this `implicit val configConvert` here? I mean - before my changes on this PR.

PR #270, opened 2022-09-05
Bundled plugins fix

wiacekm:


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
```diff
@@ -41,8 +41,12 @@ sealed trait IntelliJProvider {
 
     val targetDir = intelliJ.paths.bundledPlugins
     val archives = withParallel[Plugin, PluginArchive](allPlugins)(_.map { plugin =>
-      val file = dependencies.plugin.fetch(plugin)
-      PluginArchive(plugin, file.toExtracted)
+      val fileOpt = dependencies.plugin.fetchOpt(plugin)
```
lukaszwawrzyk, line 44: Isn't behavior the same as if you called fetch(plugin) as before?

PR #275, opened 2022-09-12
Replace IntelliJProvider.Default usages with values from config

### LINKED ISSUE #253: Replace default values of `IntelliJFixture` with loading from `reference.conf`

LukaszKontowski:
Closes #253

Each change from the list below corresponds to one commit

1. removed `IntelliJProvider.Default` as default values are loaded from `reference.conf` (please see related change nr 9.)
2. added `Config.fromReferenceConf` method
3. removed `IntelliJVersion.Latest` from scala code (read from .conf file)
4. removed `IdeProbePaths.Default` and `ResourceProvider.Default`
5. removed `Config.Empty` and `WorkspaceTemplate.Empty` defaults from IntelliJFixture
6. moved `plugins.repository.uri` default value from code to `reference.conf`
7. moved `jbr.repositories` default value from code to `reference.conf`
8.  added `lazy val Default: IntelliJFixture` to trait IdeProbeFixture to make `ide-probe` usage from code more convenient
9. refactored default values for the `IntelliJFixture` constructor (by loading from `reference.conf`)
10. added default values to the `probe.resolvers.intellij` config + a new method for users - `IntelliJResolver.fromConfig` - which uses just the values from `.conf` file by default. Also, refactored `IntelliJZipResolver.fromConfig` method to use configs from `.conf` file but with `.zip` extension ensured.
11. improved `reference.conf` readability
12. added missing description for `probe.resolvers.intellij.artifact` (PR fix)
13. made `artifact` config independent from intellij config in resolvers (PR fix)
14. Removed `probe.resolvers.artifact` config from code (PR fix)
15. Restored IntelliJProvider.Default and it's usage (PR fix)
16. Fixed user-facing comments about .dmg extension (PR fix)
17. Restored convenient Default vals (PR fix)
18. Restored ProbeDriverJavaTest with previous logic (PR fix)
19. fixed object IntelliJVersion logic for snapshot / release methods (PR fix)
20. Removed all code repetitions for the default IdeProbeConfig read from .reference.conf (PR fix)
21. Restored PluginResolver.Official in code (PR fix)
22. reverted some changes in test code (PR fix)
23. Restored JbrResolvers.official in code and made all Default values (read from config) lazy (PR fix)
24. Updated outdated code comment (format -> ext, PR fix)
25. Ranamed defaultIdeProbeConfig to defaultConfig (PR fix)
26. Changed ext field from Option[String] to String (PR fix)
27. Restored Default values in tests (PR fix)
28. restored Default vals usage in tests (PR fix)

### DIFF
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -63,72 +63,70 @@ probe {
   // `resolvers` is for `org.virtuslab.ideprobe.config.DependenciesConfig.Resolvers` - default values below
   resolvers {
 
-    intellij {
+    // `artifact` contains the name of the JetBrains product that should be used as the testing instance.
+    // For now, only IntelliJ IDEA is supported. There are two possible values:
+    // 1) "ideaIC" - for the IntelliJ IDEA Community edition  (default option)
+    // 2) "ideaIU" - for the IntelliJ IDEA Ultimate Edition
+    artifact = "ideaIC"
 
-      // `intellij.artifact` contains the name of the artifact that should be used as the IntelliJ IDEA instance.
-      // For now, there are two possible values:
-      // "ideaIC" - for the Community edition and "ideaIU" - for the Ultimate Edition
-      artifact = "ideaIC"  # or "ideaIU" - for the Ultimate Edition
-
-      // `intellij.repositories` contain patterns which should be resolved by ide-probe to download proper IntelliJ IDEA
-      // edition (Community or Ultimate) from proper intelliJ repository. Words inside square brackets
-      // are parts of the pattern that are replaced with proper values before download attempt. Current replacements
-      // are as follows:
-      //
-      // "[orgPath]" -> for now it is always replaced by "com/jetbrains/intellij"
-      // "[organisation]" -> for now it is always replaced by "com.jetbrains.intellij"
-      // "[module]" -> for now it is always replaced by "idea"
-      // "[artifact]" -> value from `probe.resolvers.intellij.artifact` - two values are valid:
-      //     "ideaIC" (community edition - default value) or "ideaIU" (ultimate edition).
-      // "[ext]" -> value from `probe.intellij.version.ext` (".zip" by default)
-      // "[revision]" -> value from `probe.intellij.version.release` - or from `probe.intellij.version.build` if `release` not found
-      // "[build]" -> value from `probe.intellij.version.build`
-      // "[version]" -> value from `probe.intellij.version.release` - or from `probe.intellij.version.build` if `release` not found
-      // "[release]" -> value from `probe.intellij.version.release` if exists. If not - "[release]" will be replaced with
-      //    the "snapshot-release" string. This replacement can be used to make sure that IntelliJ will be downloaded only
-      //    based on IntelliJVersion's "release" - and NOT based on "build".
-      //
-      // You can use a custom intellij repository pattern(s) instead of the default ones by defining your patterns
-      // in the `intellij.repositories` config.
-      //
-      // You can also add a custom intellij repository pattern(s) on top of the 6 default repositories, so that
-      // ide-probe will first try using the IntelliJs downloaded from your repositories if present. If ide-probe doesn't
-      // find the IntelliJ distribution by your patterns, it will use the default 6 repository patterns listed below.
-      // To achieve that, specify a custom repository pattern and add the `official` or `default` keyword on another line.
-      // Example below:
-      // intellij.repositories = [
-      //   "file:///"${HOME}"/.my_intellij_idea_downloads/[revision]/[artifact]-[revision][ext]",
-      //   official
-      // ]
-      //
-      // Moreover, you can add a pattern pointing to a directory of an installed IntelliJ instance instead of
-      // pattern pointing to a .zip/.dmg file. ide-probe will recognize if the pattern points to a directory with
-      // installed IntelliJ or to a .zip/.dmg file. For example, if you run the IntelliJ Plugin Verifier on your
-      // repository (https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#tasks-runpluginverifier),
-      // then you could use following configuration:
-      // intellij.repositories = [
-      //   "file:///"${HOME}"/.pluginVerifier/ides/IC-[revision]/",
-      //   official
-      // ]
-      //
-      // Such config will make ide-probe use an installed IntelliJ from local filesystem from the
-      // ~/.pluginVerifier/ides/IC-[revision]/ directory (if exists). If it does not exist under specified path,
-      // then ide-probe will download it from one of the 6 official repositories.
-      //
-      // Note: if you use the `intellij.repositories` config with a value pointing to an installed
-      // IntelliJ instance directory - then DO NOT use the `probe.intellij.path` config in the same time.
-      // The `probe.intellij.path` config takes precedence over `intellij.repositories` in such case, so the path
-      // from `intellij.repositories` would be ignored.
-      //
-      repositories = [
-        "https://download.jetbrains.com/[module]/nightly/[artifact]-[revision][ext]",
-        "https://download.jetbrains.com/[module]/[artifact]-[revision][ext]",
-        "https://download.jetbrains.com/[module]/nightly/[artifact]-[revision].portable[ext]",
-        "https://download.jetbrains.com/[module]/[artifact]-[revision].portable[ext]",
-        "https://www.jetbrains.com/intellij-repository/releases/[orgPath]/[module]/[artifact]/[revision]/[artifact]-[revision][ext]",
-        "https://www.jetbrains.com/intellij-repository/snapshots/[orgPath]/[module]/[artifact]/[revision]-EAP-SNAPSHOT/[artifact]-[revision]-EAP-SNAPSHOT[ext]"
-      ]
-    }
+    // `intellij.repositories` contain patterns which should be resolved by ide-probe to download proper IntelliJ IDEA
+    // edition (Community or Ultimate) from proper intelliJ repository. Words inside square brackets
+    // are parts of the pattern that are replaced with proper values before download attempt. Current replacements
+    // are as follows:
+    //
+    // "[orgPath]" -> for now it is always replaced by "com/jetbrains/intellij"
+    // "[organisation]" -> for now it is always replaced by "com.jetbrains.intellij"
+    // "[module]" -> for now it is always replaced by "idea"
+    // "[artifact]" -> value from `probe.resolvers.intellij.artifact` - two values are valid:
+    //     "ideaIC" (community edition - default value) or "ideaIU" (ultimate edition).
+    // "[ext]" -> value from `probe.intellij.version.ext` (".zip" by default)
+    // "[revision]" -> value from `probe.intellij.version.release` - or from `probe.intellij.version.build` if `release` not found
+    // "[build]" -> value from `probe.intellij.version.build`
+    // "[version]" -> value from `probe.intellij.version.release` - or from `probe.intellij.version.build` if `release` not found
+    // "[release]" -> value from `probe.intellij.version.release` if exists. If not - "[release]" will be replaced with
+    //    the "snapshot-release" string. This replacement can be used to make sure that IntelliJ will be downloaded only
+    //    based on IntelliJVersion's "release" - and NOT based on "build".
+    //
+    // You can use a custom intellij repository pattern(s) instead of the default ones by defining your patterns
+    // in the `intellij.repositories` config.
+    //
+    // You can also add a custom intellij repository pattern(s) on top of the 6 default repositories, so that
+    // ide-probe will first try using the IntelliJs downloaded from your repositories if present. If ide-probe doesn't
+    // find the IntelliJ distribution by your patterns, it will use the default 6 repository patterns listed below.
+    // To achieve that, specify a custom repository pattern and add the `official` or `default` keyword on another line.
+    // Example below:
+    // intellij.repositories = [
+    //   "file:///"${HOME}"/.my_intellij_idea_downloads/[revision]/[artifact]-[revision][ext]",
+    //   official
+    // ]
+    //
+    // Moreover, you can add a pattern pointing to a directory of an installed IntelliJ instance instead of
+    // pattern pointing to a .zip/.dmg file. ide-probe will recognize if the pattern points to a directory with
+    // installed IntelliJ or to a .zip/.dmg file. For example, if you run the IntelliJ Plugin Verifier on your
+    // repository (https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#tasks-runpluginverifier),
+    // then you could use following configuration:
+    // intellij.repositories = [
+    //   "file:///"${HOME}"/.pluginVerifier/ides/IC-[revision]/",
+    //   official
+    // ]
+    //
+    // Such config will make ide-probe use an installed IntelliJ from local filesystem from the
+    // ~/.pluginVerifier/ides/IC-[revision]/ directory (if exists). If it does not exist under specified path,
+    // then ide-probe will download it from one of the 6 official repositories.
+    //
+    // Note: if you use the `intellij.repositories` config with a value pointing to an installed
+    // IntelliJ instance directory - then DO NOT use the `probe.intellij.path` config in the same time.
+    // The `probe.intellij.path` config takes precedence over `intellij.repositories` in such case, so the path
+    // from `intellij.repositories` would be ignored.
+    //
+    intellij.repositories = [
+      "https://download.jetbrains.com/[module]/nightly/[artifact]-[revision][ext]",
+      "https://download.jetbrains.com/[module]/[artifact]-[revision][ext]",
+      "https://download.jetbrains.com/[module]/nightly/[artifact]-[revision].portable[ext]",
+      "https://download.jetbrains.com/[module]/[artifact]-[revision].portable[ext]",
+      "https://www.jetbrains.com/intellij-repository/releases/[orgPath]/[module]/[artifact]/[revision]/[artifact]-[revision][ext]",
+      "https://www.jetbrains.com/intellij-repository/snapshots/[orgPath]/[module]/[artifact]/[revision]-EAP-SNAPSHOT/[artifact]-[revision]-EAP-SNAPSHOT[ext]"
+    ]
 
     // `plugins.repository.uri` is the URI of the plugin repository. Use this config if you want to use a non-standard
     // plugin repository. The standard one is defined below:
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -55,12 +55,11 @@ probe {
     intellij {
       repositories = []
     }
-    // default configuration for `plugins` is Plugins(None) - adding possible parameters in comments for reference
-    plugins {
-      // `repository.uri` is the URI of the plugin repository - use this config if you want to use a non-standard
-      // plugin repository. The standard one used by ide-probe is: https://plugins.jetbrains.com/plugin/download
-      # repository.uri = ""
-    }
+
+    // `plugins.repository.uri` is the URI of the plugin repository. Use this config if you want to use a non-standard
+    // plugin repository. The standard one is defined below:
+    plugins.repository.uri = "https://plugins.jetbrains.com/plugin/download"
+
     jbr {
       repositories = []
     }
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -14,11 +14,10 @@ probe {
       build = "212.5080.55"
       release = "2021.2.1"
 
-      // `ext` stands for the file extension (file format). You can choose which extension should be downloaded.
-      // By default, ".zip" distributions of IntelliJ IDEA are downloaded. Non-default supported extension is ".dmg".
-      // Other extensions are not supported. You can choose only one extension per .conf file.
+      // `ext` stands for the file extension (file format). This config specifies which extension should be
+      // downloaded and installed. By default, ".zip" distributions of IntelliJ IDEA are downloaded. For now
+      // only ".zip" extension is supported.
       ext = ".zip"
-      # ext = ".dmg"
 
     }
 
@@ -97,8 +96,8 @@ probe {
     // ]
     //
     // Moreover, you can add a pattern pointing to a directory of an installed IntelliJ instance instead of
-    // pattern pointing to a .zip/.dmg file. ide-probe will recognize if the pattern points to a directory with
-    // installed IntelliJ or to a .zip/.dmg file. For example, if you run the IntelliJ Plugin Verifier on your
+    // pattern pointing to a .zip file. ide-probe will recognize if the pattern points to a directory with
+    // installed IntelliJ or to a .zip file. For example, if you run the IntelliJ Plugin Verifier on your
     // repository (https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#tasks-runpluginverifier),
     // then you could use following configuration:
     // intellij.repositories = [
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -60,9 +60,36 @@ probe {
     // plugin repository. The standard one is defined below:
     plugins.repository.uri = "https://plugins.jetbrains.com/plugin/download"
 
-    jbr {
-      repositories = []
-    }
+    // `jbr.repositories` contains patterns for JBR versions which should be downloaded and used for tests.
+    // ide-probe replaces words in square brackets with replacements as follows:
+    // "[major]" -> the "major" part of the JBR release's version number - e.g. "17.0.4" from "17.0.4b469.53"
+    // "[minor]" -> the "minor" part of the JBR release's version number - e.g. "b469.53" from "17.0.4b469.53"
+    // "[platform]" -> value connected to the detected OS - "linux", "osx" or "windows"
+    //
+    // You can define your own `jbr.repositories` if you want to use a non-standard JBR repository pattern
+    // instead of the 2 standard ones defined below.
+    //
+    // You can also add a custom JBR repository on top of the 2 default repositories, so that ide-probe will
+    // first try using the JBRs downloaded from your repository if present. If ide-probe doesn't find the JBR
+    // package in your repository, it will use the default 2 repositories listed below. To achieve that, specify
+    // your custom repository pattern and add the `official` or `default` keyword on another line. Example below:
+    // jbr.repositories = [
+    //   "https://repositories.my-company.com/intellij-jbr/jbr-[major]-[platform]-x64-b[minor].tar.gz",
+    //   official
+    // ]
+    //
+    // The `jbr_dcevm` pattern applies only to versions of IntelliJ older than 2022.2 release, whereas
+    // `jbr` pattern applies to newer versions where DCEVM is bundled by default.
+    // Another important note: currently we are using only JBR packages with `x64` architecture support. During tests
+    // it turned out that using `aarch64` packages on processors with `aarch64` architecture (Mac M1 Pro chip) results
+    // in "Cannot load JVM bundle:" error. For some reason only `x64` packages work well (both for `x64` and `aarch64`
+    // processor architectures). Hence `x64` should stay hardcoded in the `jbr.repositories` elements.
+    //
+    jbr.repositories = [
+      "https://cache-redirector.jetbrains.com/intellij-jbr/jbr_dcevm-[major]-[platform]-x64-b[minor].tar.gz",
+      "https://cache-redirector.jetbrains.com/intellij-jbr/jbr-[major]-[platform]-x64-b[minor].tar.gz"
+    ]
+
     retries = 0
   }
 
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -63,58 +63,63 @@ probe {
   // `resolvers` is for `org.virtuslab.ideprobe.config.DependenciesConfig.Resolvers` - default values below
   resolvers {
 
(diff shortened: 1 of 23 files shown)


core/driver/sources/src/main/resources/reference.conf
```diff
@@ -52,18 +62,108 @@ probe {
...
-      // plugin repository. The standard one used by ide-probe is: https://plugins.jetbrains.com/plugin/download
-      # repository.uri = ""
-    }
-    jbr {
-      repositories = []
+
+      // `intellij.artifact` contains the name of the artifact that should be used as the IntelliJ IDEA instance.
+      // For now, there are two possible values:
+      // "ideaIC" - for the Community edition and "ideaIU" - for the Ultimate Edition
+      artifact = "ideaIC"  # or "ideaIU" - for the Ultimate Edition
```
PawelLipski, line 71: There are ofc other JetBrains products with their respective artifact names... but actually, does ide-probe support anything beyond IntelliJ currently? 🤔

PawelLipski (approved):
:benissimo:, but leaving to Łukasz for the final review ;D

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: Only IntelliJ IDEA is supported for now. Maybe move from `probe.resolvers.intellij.artifact` to `probe.resolvers.artifact` to support possibility of adding other JetBrains products??

PawelLipski: Huh such moving (renaming) of the key would be a breaking change right? 🤔 I'd leave as is, to be re-considered once the need for support of other IDEs is really needed

PawelLipski: E.g. as for git-machete-intellij-plugin, I haven't got any reports that the plugin, once thoroughly tested on IntelliJ, ever fails on other IDEs

LukaszKontowski: Ok, but this config: `probe.resolvers.intellij.artifact` is a new one anyway. I'm adding it in this PR... So no breaking current code from master branch.

PawelLipski: @lukaszwawrzyk are you aware of any corner case where a plugin would NOT work on the non-IntelliJ IDEs even if it works on IntelliJ (ofc assuming all dependencies from plugin.xml are satisfied by that IDE, but that's enforced statically anyway &mdash; you won't even install the plugin if such dependencies are missing).

PawelLipski: > Ok, but this config: probe.resolvers.intellij.artifact is a new one anyway. I'm adding it in this PR... So no breaking of current code from master branch.

Aaa okay, so indeed, let's make it future-proof by avoiding explicit references to `intellij`

LukaszKontowski: Ok, moved from `probe.resolvers.intellij.artifact` to `probe.resolvers.artifact` as discussed

lukaszwawrzyk: @PawelLipski I recall for example that PyCharm does not behave or look the same in all situations as IntelliJ with Python plugin.

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -52,18 +62,106 @@ probe {
...
-    }
-    jbr {
-      repositories = []
-    }
+
+    // `artifact` contains the name of the JetBrains product that should be used as the testing instance.
+    // For now, only IntelliJ IDEA is supported. There are two possible values:
+    // 1) "ideaIC" - for the IntelliJ IDEA Community edition  (default option)
+    // 2) "ideaIU" - for the IntelliJ IDEA Ultimate Edition
+    artifact = "ideaIC"
```
lukaszwawrzyk, line 70: You could just keep the artifact in patterns as it was. It is beyond scope of this PR. I am not sure if artifact is best name in the config (it makes sense in resolver patterns though). Also scope here is too wide. There is data for IJ and data for plugin, also for Jbr, and. this is just some dangling global artifact. We would probably add a new grouping key or make an aware breaking change to generalize it further. I don't think it is worth doing now and I don't want it to end up with a bad design.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IntelliJFixture.scala
```diff
@@ -15,7 +16,7 @@ import org.virtuslab.ideprobe.ide.intellij.RunningIde
 
 final case class IntelliJFixture(
     workspaceProvider: WorkspaceProvider = WorkspaceTemplate.Empty,
-    intelliJProvider: IntelliJProvider = IntelliJProvider.Default,
+    intelliJProvider: IntelliJProvider = getIntelliJProvider(readIdeProbeConfig(Config.fromReferenceConf, ConfigRoot)),
```
lukaszwawrzyk, line 19: Library user should still be able to say `IntelliJProvider.Default`

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
```diff
@@ -15,6 +33,7 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
       "orgPath" -> "com/jetbrains/intellij",
       "module" -> "idea",
       "artifact" -> artifact,
+      "ext" -> version.ext.get, // .get will be OK since we have `format = ".zip"` in the reference.conf file
```
lukaszwawrzyk, line 36: why it is optional, if it is ok to call 'get'?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
```diff
@@ -31,14 +33,12 @@ final case class IntelliJVersion(build: String, release: Option[String]) {
 object IntelliJVersion {
   implicit val configConvert: ConfigConvert[IntelliJVersion] = deriveConvert[IntelliJVersion]
 
-  // TODO(#253): replace Latest with reading from reference.conf
-  val Latest: IntelliJVersion = release("2021.2.1", "212.5080.55")
-
+  // ext = Some(configConvert.map(_.ext).toString) is a convenient way to receive `probe.intellij.version.ext` value
   def snapshot(build: String): IntelliJVersion = {
-    IntelliJVersion(build, None)
+    IntelliJVersion(build = build, release = None, ext = Some(configConvert.map(_.ext).toString))
```
lukaszwawrzyk, line 38: If you need to read base IJVersion from config, you can just use  something like `Latest.copy(build = build, release = release)` in both methods. And read `Latest` properly (from reference conf)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJZipResolver.scala
```diff
@@ -37,17 +38,10 @@ object IntelliJZipResolver extends IntelliJResolver {
...
-      AlternativeIntelliJZipResolver.community,
-      NightlyIntelliJZipResolver.community
-    )
-    val fromConfig = config.repositories
-      .flatMap(pattern =>
-        if (Set("official", "default").contains(pattern.toLowerCase)) official
-        else Seq(IntelliJPatternResolver(pattern).community)
-      )
-    if (fromConfig.isEmpty) official else fromConfig
+  // this method enables downloading .zip IntelliJ even if `probe.intellij.version.ext` is ".dmg"
```
lukaszwawrzyk, line 41: Why is it needed? Is installing IJ from ".dmg" supported or not?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/ResourceProvider.scala
```diff
@@ -21,8 +21,6 @@ object ResourceProvider {
     new Cached(paths.cache, retries)
   }
 
-  val Default = new Cached(IdeProbePaths.Default.cache, 0)
```
lukaszwawrzyk, line 24: can you not delete these convenient `Default` vals that helps constructing other classes from code?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -73,7 +73,8 @@ sealed abstract class InstalledIntelliJ(root: Path, probePaths: IdeProbePaths, c
 
   private val executable: Path = {
     val content = {
-      val macOsLauncher = paths.root.resolve("MacOS").resolve("idea")
+      val macOsLauncher =
+        paths.root.resolve("MacOS").resolve("idea") // TODO - move "idea" to driver.launch.command ... ?
```
lukaszwawrzyk, line 77: don't add todo comments especially if you are unsure. Open an issue and describe the problem.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
```diff
@@ -161,33 +153,13 @@ final case class IntelliJFactory(
 }
 
 object IntelliJProvider {
-  val Default: IntelliJFactory =
```
lukaszwawrzyk, line 164: Please keep `val Default: IntelliJFactory` (keep the type of it as well)

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/SingleRunFixtureTest.scala
```diff
@@ -56,7 +57,7 @@ final class SingleRunFixtureTest extends IdeProbeFixture with WorkspaceFixture w
 
   @Test
   def removesDirectoriesEvenAfterFailureToRunIntelliJ(): Unit = {
-    val intelliJFixture = IntelliJFixture().withAfterIntelliJInstall((_, intellij) =>
```
lukaszwawrzyk, line 59: Keep the original usage

core/driver/tests/src/test/java/org/virtuslab/ideprobe/ProbeDriverJavaTest.java
```diff
@@ -1,40 +1,23 @@
...
 import static org.virtuslab.ideprobe.WaitLogic.emptyBackgroundTasks;
-import static org.virtuslab.ideprobe.WorkspaceTemplate.FromResource;
 import static org.virtuslab.ideprobe.wait.WaitLogicFactory.*;
-import static scala.collection.JavaConverters.collectionAsScalaIterable;
 import static scala.collection.JavaConverters.seqAsJavaList;
 
 public class ProbeDriverJavaTest {
     @Test
     public void openProject() {
-        var workspaceProvider = new FromResource("gradle-project");
```
lukaszwawrzyk, line 28: I am not sure about this change. You could add it as another test if you want. 
Both old and new version show how to setup IJFixture with a custom workspace from Java in 2 different ways (from code and from config).

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: Ok, removed `probe.resolvers.artifact` config from code. `[artifact]` replacement should work just like before

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJZipResolver.scala
LukaszKontowski: Looks like installing from .dmg is not supported in reality. Created an issue: https://github.com/VirtusLab/ide-probe/issues/281
If we add the support for `.dmg` in the future, then it would be needed. Let me fix user-facing comments so that it will be obvious, that for now we do not support `.dmg` extensions.

LukaszKontowski: Fixed user-facing comments about .dmg extension

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/ResourceProvider.scala
LukaszKontowski: Restored convenient Default vals

PawelLipski (approved):

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJZipResolver.scala
LukaszKontowski: Support for installing from `.dmg` should be available after merging https://github.com/VirtusLab/ide-probe/pull/283

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IdeProbePaths.scala
```diff
@@ -17,8 +17,8 @@ case class IdeProbePaths(
 
 object IdeProbePaths {
   val Default: IdeProbePaths = {
-    // TODO: replace None parameters with loading defaults from reference.conf
-    from(PathsConfig(None, None, None, None, None, None, None))
+    val config = IntelliJFixture.readIdeProbeConfig(Config.fromReferenceConf, "probe")
```
lukaszwawrzyk, line 20: we shouldn't need to repeat this "probe" anywhere. 
Also `IntelliJFixture.readIdeProbeConfig(Config.fromReferenceConf, "probe")` could be stored as a lazy val to avoid reading it multiple times.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
```diff
@@ -3,7 +3,13 @@ package org.virtuslab.ideprobe.dependencies
...
 import pureconfig.generic.semiauto.deriveConvert
 
-final case class IntelliJVersion(build: String, release: Option[String]) {
+import org.virtuslab.ideprobe.Config
+import org.virtuslab.ideprobe.IntelliJFixture
+import org.virtuslab.ideprobe.config.IntellijConfig
+
+// the `ext` field (file extension) must be an Option[String] and not a String for the ExistingIntelliJ usage
+// inside the `org.virtuslab.ideprobe.dependencies.IntelliJVersionResolver` object
+final case class IntelliJVersion(build: String, release: Option[String], ext: Option[String]) {
```
lukaszwawrzyk, line 12: Maybe it doesn't need to be optional and could be just ignored by existing IJ?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
```diff
@@ -31,14 +37,18 @@ final case class IntelliJVersion(build: String, release: Option[String]) {
 object IntelliJVersion {
   implicit val configConvert: ConfigConvert[IntelliJVersion] = deriveConvert[IntelliJVersion]
 
-  // TODO(#253): replace Latest with reading from reference.conf
-  val Latest: IntelliJVersion = release("2021.2.1", "212.5080.55")
+  private lazy val defaultConfig = IntelliJFixture
```
lukaszwawrzyk, line 40: instead of repeating this code, it could just be in IntelliJFixture, computed once.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJZipResolver.scala
```diff
@@ -37,17 +38,10 @@ object IntelliJZipResolver extends IntelliJResolver {
...
-      NightlyIntelliJZipResolver.community
-    )
-    val fromConfig = config.repositories
-      .flatMap(pattern =>
-        if (Set("official", "default").contains(pattern.toLowerCase)) official
-        else Seq(IntelliJPatternResolver(pattern).community)
-      )
-    if (fromConfig.isEmpty) official else fromConfig
+  // this method enables downloading .zip IntelliJ even if `probe.intellij.version.ext` value is not ".zip"
+  def fromConfig(config: DependenciesConfig.Resolvers): Seq[DependencyResolver[IntelliJVersion]] = {
```
lukaszwawrzyk, line 42: This sounds hacky. What would happen if we just don't replace it? Shouldn't it just fail if someone provides unsupported extension?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/JbrResolvers.scala
```diff
@@ -11,29 +11,14 @@ import org.virtuslab.ideprobe.dependencies.Dependency.Missing
...
-      .flatMap(pattern =>
-        if (Set("official", "default").contains(pattern.toLowerCase)) official
-        else Seq(JbrPatternResolver(pattern))
-      )
-    if (fromConfig.isEmpty) official else fromConfig
-  }
+  def fromConfig(config: DependenciesConfig.Jbr): Seq[DependencyResolver[Path]] =
+    config.repositories.flatMap { pattern =>
+      if (Set("official", "default").contains(pattern.toLowerCase)) {
+        val probeConfigFromReference = IntelliJFixture.readIdeProbeConfig(Config.fromReferenceConf, "probe")
```
lukaszwawrzyk, line 17: another instance of repeated code, please make sure that you track all of these and replace them.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/PluginResolver.scala
```diff
@@ -5,16 +5,13 @@ import org.virtuslab.ideprobe.dependencies.Plugin._
 import org.virtuslab.ideprobe.error
 
 object PluginResolver {
-  // TODO (#253): replace Official with a default value in reference.conf - `probe.resolvers.plugins.repository.uri`
-  val Official: DependencyResolver[Plugin] = PluginResolver("https://plugins.jetbrains.com/plugin/download")
```
lukaszwawrzyk, line 9: Again, please do not drop useful apis. Read it from config if necessary, but keep the val.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJProviderTest.scala
```diff
@@ -26,8 +33,8 @@ final class IntelliJProviderTest {
 
       // then
       assert(
-        intelliJVersion.build == IntelliJProvider.Default.version.build,
-        s"Expected ${IntelliJProvider.Default.version}, but got $intelliJVersion."
+        intelliJVersion.build == defaultIntellijProvider.version.build,
```
lukaszwawrzyk, line 36: this change shouldn't be necessary

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJProviderTest.scala
```diff
@@ -56,7 +63,7 @@ final class IntelliJProviderTest {
 
   @Test
   def shouldInstallIntellijFromExtractedRepository(): Unit = givenInstalledIntelliJ { installationRoot =>
-    val build = IntelliJVersion.Latest.build
+    val build = probeConfig.intellij.asInstanceOf[IntellijConfig.Default].version.build
```
lukaszwawrzyk, line 66: keep original code

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJZipResolver.scala
LukaszKontowski: My changes slightly change the logic used in `IntelliJProvider.from` so that:
- it no longer uses `IntelliJZipResolver.fromConfig`
- it uses `IntelliJResolver.fromConfig` instead - as the default method called when reading from config

So - `IntelliJZipResolver.fromConfig` will be a method that is not used by default when reading from config. It still could be used directly from code. It could be useful in some scenarios like below:
- if somebody uses `.dmg` as `probe.intellij.version.ext` (which will work after merging https://github.com/VirtusLab/ide-probe/pull/283 ) - but wants to use `.zip` in some specific scenario
- if somebody uses an Existing IntelliJ instead of `.zip` by default  - but wants to use `.zip` in some specific scenario

So, if someone provides unsopported extension - it will fail. But user's config will not go through `IntelliJZipResolver.fromConfig` but through `IntelliJResolver.fromConfig` (which is a more general method - could be used for more than `.zip` extension)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/PluginResolver.scala
LukaszKontowski: sure, will be fixed in upcomming commit

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
LukaszKontowski: Hm 🤔 What exactly do you have in mind? I don't understand to be honest. The thing is - `IntelliJVersionResolver.version` - used only for `ExistingIntelliJ` - returns an instance of `IntelliJVersion`. And after my changes - `IntelliJVersion` will have the additional `ext` field. But this in fact is valid and makes sense only for using `IntellijConfig.Default` - but makes no sense for `IntellijConfig.Existing`.

What do you suggest?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
```diff
@@ -15,6 +31,7 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
...
```
lukaszwawrzyk, line 34: this comment is outdated, I think it is `ext = "zip"`. Make sure that dot is handled/expected consistently in all places.

LukaszKontowski: ok, let me fix the comment - in fact we have `ext = ".zip"` with a dot

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
lukaszwawrzyk: I understand this. This is a bit ugly though to call .get because it is valid in specific context. I was considering using String rather than Option[String], and just put a special value for existing case or ignore it in that context.

Another valid approach would be to store "ext" elsewhere or have some ADT, so that version for downloadable IJ has this field, but version for existing IJ does not have this field. It might be complicated to integrate it with rest of the code nicely though.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IntelliJFixture.scala
```diff
@@ -129,17 +129,20 @@ final case class IntelliJFixture(
 
 object IntelliJFixture {
   private val ConfigRoot = "probe"
+  lazy val defaultIdeProbeConfig: IdeProbeConfig = readIdeProbeConfig(Config.fromReferenceConf, "probe")
```
lukaszwawrzyk, line 132: can we just call it `default config`? Also use `ConfigRoot` rather than explicitly "probe" string.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJZipResolver.scala
lukaszwawrzyk: Ah, I see. It is an extra method to load resolvers from config, but turn them into zip resolvers. Makes sense.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/IntelliJFixture.scala
LukaszKontowski: fixed as suggested

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
LukaszKontowski: fixed as suggested

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJProviderTest.scala
```diff
@@ -11,13 +11,18 @@ import org.junit.runners.JUnit4
...
+import org.virtuslab.ideprobe.IdeProbePaths
 import org.virtuslab.ideprobe.IntelliJFixture
 import org.virtuslab.ideprobe.ide.intellij.IntelliJProvider
 
 @RunWith(classOf[JUnit4])
 final class IntelliJProviderTest {
   private implicit val ec: ExecutionContext = ExecutionContext.fromExecutorService(Executors.newCachedThreadPool())
 
+  private val probeConfig = IntelliJFixture.defaultIdeProbeConfig
+  private val defaultIntellijProvider = IntelliJProvider
```
lukaszwawrzyk, line 23: isn't that just IntelliJProvider.Default

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/OfficialResolversTest.scala
```diff
@@ -7,30 +7,69 @@ import org.junit.Test
...
+      val resolved = dependencyResolver.resolve(probeConfig.intellij.asInstanceOf[IntellijConfig.Default].version)
+      resolved.asInstanceOf[Dependency.Artifact].uri.toString.endsWith("-EAP-SNAPSHOT.zip")
+    }
+    assertTrue(snapshotResolverOption.nonEmpty)
+  }
+
   @Test
   def resolvesBuildToExistingArtifact(): Unit = {
-    val version = IntelliJVersion.Latest
+    val version = defaultProbeConfig.intellij.asInstanceOf[IntellijConfig.Default].version
```
lukaszwawrzyk, line 50: keep original code

core/driver/tests/src/test/scala/org/virtuslab/ideprobe/ProbeDriverTest.scala
```diff
@@ -26,11 +26,14 @@ import org.virtuslab.ideprobe.robot.RobotPluginExtension
 
 @RunWith(classOf[JUnit4])
 final class ProbeDriverTest extends IdeProbeFixture with Assertions with RobotPluginExtension {
-  private val intelliJProvider = IntelliJProvider.Default
```
lukaszwawrzyk, line 29: you should be able to keep the original code

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
LukaszKontowski: Changed ext field from Option[String] to String

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/IntelliJProviderTest.scala
LukaszKontowski: yes, changed to IntelliJProvider.Default

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/OfficialResolversTest.scala
LukaszKontowski: fixed as suggested

core/driver/tests/src/test/scala/org/virtuslab/ideprobe/ProbeDriverTest.scala
LukaszKontowski: fixed as suggested

LukaszKontowski:
https://github.com/VirtusLab/ide-probe/actions/runs/3112925591/jobs/5046909976 - this looks like a random failure not related to my commits. Let me push again to restart the CI jobs.

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/OfficialResolversTest.scala
```diff
@@ -27,10 +53,23 @@ final class OfficialResolversTest {
     verify(uri)
   }
 
+  @Test
+  def resolvesBuildToExistingArtifactFromConfig(): Unit = {
+    val intellijVersion = defaultProbeConfig.intellij.asInstanceOf[IntellijConfig.Default].version
```
lukaszwawrzyk, line 58: isn't t just IntelliJVersion.Latest?

lukaszwawrzyk (approved):

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/OfficialResolversTest.scala
```diff
@@ -27,10 +53,23 @@ final class OfficialResolversTest {
...
+      resolved.asInstanceOf[Dependency.Artifact].uri.toString.contains(officialReleasesRepositoryURL)
+    }.get
+    verify(releaseResolver.resolve(intellijVersion))
+  }
+
   @Test
   def resolvesPluginToExistingArtifact(): Unit = {
     val plugin = Plugin("org.intellij.scala", "2020.2.7")
-    val uri = PluginResolver.Official.resolve(plugin)
+    val uri = PluginResolver.fromConfig(defaultProbeConfig.resolvers.plugins).resolve(plugin)
```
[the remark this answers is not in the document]
LukaszKontowski: done

core/driver/sources/src/test/scala/org/virtuslab/ideprobe/dependencies/OfficialResolversTest.scala
LukaszKontowski: yes, changed as suggested

lukaszwawrzyk (approved):

PR #280, opened 2022-09-14
add automated intellij version bump

### LINKED ISSUE #256: Add automated intelliJ version bump
Should be done after completing https://github.com/VirtusLab/ide-probe/issues/253 - so that we will not have version hardcoded in Scala code like it is now:
```
val Latest = release("2021.2.1", "212.5080.55")
```

IntelliJ version from `reference.conf` file should be updated automatically. Could be done by using some github bot user and finding newest release from https://www.jetbrains.com/intellij-repository/releases/ - or maybe some other, smarter way.

LukaszKontowski:
Closes #256
Added scheduled job that would run daily (later to be changed to once a week) to search for new intelliJ release and create a PR if a newer release was found.

### DIFF
--- a/.github/workflows/update-intellij-version.yml
+++ b/.github/workflows/update-intellij-version.yml
@@ -8,26 +8,39 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2.3.4
-        # with:
-        #   token: 'add_secret_variable_with_bot's_personal_access_token_here'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Update default IntelliJ version
         run: |
           sbt "update-intellij / run --java-home /usr/local/openjdk-11"
-          # git config user.name BOT_USER_NAME
-          # git config user.email BOT_USER_EMAIL
-          # git add .
-          ## `cut -c 14-` extracts intellij version from line starting with "latestStable=" (like "latestStable=2021.2.1")
-          ## // TODO - cope with situations where there is nothing to commit ...
-          # git commit -m "updated default IntelliJ version to $INTELLIJ_VERSION"
-          # git push
-          # echo "Adding comment to the pull request..."
-          # curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg
-          # echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] \
-          #   https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null
-          # apt update
-          # apt install gh -y
-          # echo $BOT_TOKEN | gh auth login --with-token // TODO set BOT_TOKEN
-          # PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
-          # COMMENT_BODY="" // TODO ADD COMMENT - MENTION https://plugins.jetbrains.com/plugin/1347-scala/versions
-          # gh pr comment $PR_NUMBER --body $COMMENT_BODY 
-          
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add .
+          if [[ -n $(git status | grep 'modified') ]]
+          then
+            RELEASE_LINE=$(grep 'release = \"' core/driver/sources/src/main/resources/reference.conf)
+            INTELLIJ_VERSION=$(awk -F'\"|\"' '{print $2}' <<< "$RELEASE_LINE")
+            git commit -m "updated default IntelliJ version to $INTELLIJ_VERSION"
+            COMMIT_SHA=$(git rev-parse HEAD)
+            git fetch
+            git checkout $GITHUB_BASE_REF
+            BRANCH_NAME="update_intellij_version_to_$INTELLIJ_VERSION"
+            git branch $BRANCH_NAME
+            git checkout $BRANCH_NAME
+            git cherry-pick $COMMIT_SHA
+            git push --set-upstream origin $BRANCH_NAME
+            curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg
+            echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] \
+              https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+            apt update
+            apt install gh -y
+            echo $GITHUB_TOKEN | gh auth login --with-token
+            PR_TITLE="Update IntelliJ version to $INTELLIJ_VERSION"
+            SCALA_PLUGIN_FILE_URL="https://github.com/VirtusLab/ide-probe/blob/master/extensions/scala/tests/src/test/resources/scala.conf"
+            SCALA_PLUGIN_VERSIONS_URL="https://plugins.jetbrains.com/plugin/1347-scala/versions"
+            PR_BODY="Updates the default IntelliJ Version to the newest release. \
+              Please, check currently used version of the scala plugin: $SCALA_PLUGIN_FILE_URL \
+              and verify if it is compatible with the new intelliJ IDEA version ($INTELLIJ_VERSION) here: $SCALA_PLUGIN_VERSIONS_URL. \
+              If the version of the scala plugin used in $SCALA_PLUGIN_FILE_URL is NOT compatible - please update it in a separate commit."
+            gh pr create --title $PR_TITLE --body $PR_BODY
+          fi
--- a/.github/workflows/update-intellij-version.yml
+++ b/.github/workflows/update-intellij-version.yml
@@ -1,0 +1,33 @@
+name: Update IntelliJ Version
+
+on:
+  pull_request:
+
+jobs:
+  bump-intellij-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+        # with:
+        #   token: 'add_secret_variable_with_bot's_personal_access_token_here'
+      - name: Update default IntelliJ version
+        run: |
+          sbt "update-intellij / run --java-home /usr/local/openjdk-11"
+          # git config user.name BOT_USER_NAME
+          # git config user.email BOT_USER_EMAIL
+          # git add .
+          ## `cut -c 14-` extracts intellij version from line starting with "latestStable=" (like "latestStable=2021.2.1")
+          ## // TODO - cope with situations where there is nothing to commit ...
+          # git commit -m "updated default IntelliJ version to $INTELLIJ_VERSION"
+          # git push
+          # echo "Adding comment to the pull request..."
+          # curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg
+          # echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] \
+          #   https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+          # apt update
+          # apt install gh -y
+          # echo $BOT_TOKEN | gh auth login --with-token // TODO set BOT_TOKEN
+          # PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
+          # COMMENT_BODY="" // TODO ADD COMMENT - MENTION https://plugins.jetbrains.com/plugin/1347-scala/versions
+          # gh pr comment $PR_NUMBER --body $COMMENT_BODY 
+          
--- a/.github/workflows/update-intellij-version.yml
+++ b/.github/workflows/update-intellij-version.yml
@@ -1,47 +1,33 @@
 name: Update IntelliJ Version
 
 on:
-  pull_request:
+  schedule:
+    - cron: '0 1 * * 1'
 
 jobs:
   bump-intellij-version:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2.3.4
+      - uses: actions/checkout@v3.0.2
       - name: Update default IntelliJ version
         run: |
           sbt "update-intellij / run --java-home /usr/local/openjdk-11"
-          git config user.name github-actions
-          git config user.email github-actions@github.com
-          git add .
-          if [[ -n $(git status | grep 'modified') ]]
-          then
-            RELEASE_LINE=$(grep 'release = \"' core/driver/sources/src/main/resources/reference.conf)
-            INTELLIJ_VERSION=$(awk -F'\"|\"' '{print $2}' <<< "$RELEASE_LINE")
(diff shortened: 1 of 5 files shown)


ci/update-intellij/src/main/resources/intellijVersion.properties
```diff
@@ -0,0 +1,4 @@
+#Thu Aug 18 00:07:22 UTC 2022
+# Values below must match intellij version values from the reference.conf file. DO NOT CHANGE MANUALLY.
```
PawelLipski, line 2: Huh ideally I'd avoid repetition if possible (i.e. keep HOCON as the sole source of truth; best create a small .conf file with just the values for IntelliJ versions, that'll be easier to manage).

LukaszKontowski: You are right - unnecessary file deleted: https://github.com/VirtusLab/ide-probe/pull/280/commits/1df0b14ce3abda78a78a8f5ae7b4b519c32f759d

.github/workflows/update-intellij-version.yml
```diff
@@ -0,0 +1,46 @@
...
+      - uses: actions/checkout@v2.3.4
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Update default IntelliJ version
+        run: |
+          sbt "update-intellij / run --java-home /usr/local/openjdk-11"
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add .
+          if [[ -n $(git status | grep 'modified') ]]
```
PawelLipski, line 19: https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/.circleci/config.yml#L35

1. `git status` is a [**porcelain** command](https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain). While it's probably correct what you do here, it's still preferred to use **plumbing** commands in automated setups like CI or scripts. E.g. `git machete` doesn't use the porcelain `git branch` under the hood, but plumbing/lower-level `git for-each-ref` instead.
2. Subjective, but I prefer doing `if ! ...; then exit; fi; .........` than `if ...; then .........; fi`; handling indent levels is tricky enough in YAML, even without language injections

.github/workflows/update-intellij-version.yml
```diff
@@ -0,0 +1,46 @@
...
+
+jobs:
+  bump-intellij-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Update default IntelliJ version
+        run: |
```
PawelLipski, line 14:
```suggestion
        # language=sh
        run: |
```
This should highlight the script as a shell script injection in the IntelliJ (assuming IntelliLang plugin is enabled)

PawelLipski: (thus also highlighting certain problems, like quoting issues)

.github/workflows/update-intellij-version.yml
```diff
@@ -0,0 +1,46 @@
...
+          sbt "update-intellij / run --java-home /usr/local/openjdk-11"
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add .
```
PawelLipski, line 18: `git add .` and `git add -A .` are insecure in automated scenarios, as they might end up committing & pushing secret files. Probably not a concern for this specific CI pipeline in this specific repo, but better to use `git add -u .` (only stage modified & removed files, not new files) or `git commit -a ...`

.github/workflows/update-intellij-version.yml
```diff
@@ -0,0 +1,47 @@
...
+            curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg
+            echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] \
+              https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+            apt update
+            apt install gh -y
+            gh repo fork --remote=true
+            BRANCH_NAME="update_intellij_version_to_$INTELLIJ_VERSION"
+            git branch $BRANCH_NAME
+            git checkout $BRANCH_NAME
+            git cherry-pick $COMMIT_SHA
```
PawelLipski, line 36: Hmm cherry-pick? 🍒 ⛏️   can the changes be just committed on the right branch from the start?  🤔

ci/update-intellij/src/main/scala/org.virtuslab.ideprobe/BumpIntelliJ.scala
```diff
@@ -0,0 +1,100 @@
...
+        if (line.contains(intellijBuildNumberFromReferenceConf))
+          line.replace(intellijBuildNumberFromReferenceConf, newestStableBuildNumber)
+        else line
+      }
+      .foreach(x => w.println(x))
+    w.close()
+    Files.move(newFile.toPath, oldFile.toPath, REPLACE_EXISTING)
+  }
+
+  private def getValueFromReferenceConf(charsUntilValue: String): String = Source
```
PawelLipski, line 83: Maybe just read the file with Typesafe Config library? 🤔

LukaszKontowski: I would keep the logic it as it is for one reason - my current solution provides the same extracting logic both in `ci/update-intellij/src/main/scala/org.virtuslab.ideprobe/BumpIntelliJ.scala` and in `.github/workflows/update-intellij-version.yml`. We have:
- `private val intellijReleaseFromReferenceConf = getValueFromReferenceConf("release = \"")` in scala code
- `RELEASE_LINE=$(grep 'release = \"' core/driver/sources/src/main/resources/reference.conf)` in the script
which means - in both cases we are searching for `release = \"`.
If we change the scala logic to use Typesafe Config tricks (instead of my not-so-elegant logic) - then there will be possible scenarios, where someone changes the style in `reference.conf` - so that scala logic works fine, but the logic from script does **not** find the value...

If I should switch to use Typesafe Config library - then also the script should somehow get the same value that was extracted in scala code (some temporary file usage maybe?)

.github/workflows/update-intellij-version.yml
```diff
@@ -0,0 +1,34 @@
+name: Update IntelliJ Version
+
+on:
+  schedule:
+    - cron: '0 1 * * 1'
```
PawelLipski, line 5: Pls document the meaning of the cron in human-readable form, as in https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/.circleci/config.yml#L302

.github/workflows/update-intellij-version.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  bump-intellij-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.0.2
+      - name: Update default IntelliJ version
+        run: |
+          sbt "update-intellij / run --java-home /usr/local/openjdk-11"
```
PawelLipski, line 14: What `java-home` is assumed by default? 🤔

ci/update-intellij/src/main/scala/org.virtuslab.ideprobe/BumpIntelliJ.scala
PawelLipski: Okay, makes sense... leaving up to Łukasz W for the final review

.github/workflows/update-intellij-version.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Prepare variables for PR
+        # language=sh
+        run: |
+          release_line=$(grep 'release = \"' core/driver/sources/src/main/resources/reference.conf)
+          intellij_version=$(awk -F'\"|\"' '{print $2}' <<< "$release_line")
+          echo "BRANCH_NAME=update_intellij_version_to_$intellij_version" >> $GITHUB_ENV
```
PawelLipski, line 20: What if a branch of this name (and/or a PR for such branch) already exists? 🤔  See the use of `git ls-remote` in git-machete-intellij-plugin

.github/workflows/update-intellij-version.yml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+      - uses: actions/checkout@v3.0.2
+      - name: Update default IntelliJ version
+        run: |
+          sbt "update-intellij / run --java-home /usr/local/openjdk-11"
+      - name: Prepare variables for PR
+        # language=sh
+        run: |
+          release_line=$(grep 'release = \"' core/driver/sources/src/main/resources/reference.conf)
+          intellij_version=$(awk -F'\"|\"' '{print $2}' <<< "$release_line")
```
PawelLipski, line 19: Or even simpler
```suggestion
          intellij_version=$(cut -d'"' -f2 <<< "$release_line")
```

.github/workflows/update-intellij-version.yml
PawelLipski: Also, nit:
```suggestion
          echo "BRANCH_NAME=update_intellij_version/$intellij_version" >> $GITHUB_ENV
```
The `general_prefix/specific_thing` is a well-established convention for git branch names

.github/workflows/update-intellij-version.yml
```diff
@@ -0,0 +1,34 @@
...
+          echo "BRANCH_NAME=update_intellij_version_to_$intellij_version" >> $GITHUB_ENV
+          echo "PR_TITLE=Update IntelliJ version to $intellij_version" >> $GITHUB_ENV
+          scala_plugin_file_url="https://github.com/VirtusLab/ide-probe/blob/master/extensions/scala/tests/src/test/resources/scala.conf"
+          scala_plugin_versions_url="https://plugins.jetbrains.com/plugin/1347-scala/versions"
+          echo "PR_BODY=Updates the default IntelliJ Version to the newest release. \
+            Please, check currently used version of the scala plugin: $scala_plugin_file_url \
+            and verify if it is compatible with the new intelliJ IDEA version ($intellij_version) here: $scala_plugin_versions_url. \
+            If the version of the scala plugin used in $scala_plugin_file_url is NOT compatible - please update it in a separate commit." >> $GITHUB_ENV
+      - name: Create Pull Request
```
PawelLipski, line 28: > What if a branch of this name (and/or a PR for such branch) already exists?

Is this handled automagically by this action?

.github/workflows/update-intellij-version.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: peter-evans/create-pull-request@v4
+        with:
+          branch: ${{ env.BRANCH_NAME }}
+          base: "master"
+          title: ${{ env.PR_TITLE }}
```
PawelLipski, line 33: Since you're **not** committing explicitly, I'm assuming that the action performs the commit, filling up the commit message from `PR_TITLE` right?

.github/workflows/update-intellij-version.yml
LukaszKontowski: I'm not sure, but it looks like it might have such powers: https://github.com/peter-evans/create-pull-request#-create-pull-request
image

Anyway - let me test this scenario on my testing repo

.github/workflows/update-intellij-version.yml
LukaszKontowski: For now the default JDK version on `ubuntu-latest` is:
`openjdk version "11.0.16.1" 2022-08-12`
but I assume that it could change in time.

.github/workflows/update-intellij-version.yml
LukaszKontowski: For me both expressions are not simple to be honest (you have to study manuals of `awk` or `cut` to understand, why it gets the desired value from line...) 😄  But maybe you are right - the `cut` command looks less scary. Let me change that.

.github/workflows/update-intellij-version.yml
LukaszKontowski: No, the commit message by default is `[create-pull-request] automated change`. Let me change that by adding the `commit-message` action input in upcomming commit.

.github/workflows/update-intellij-version.yml
LukaszKontowski: ok, looks like there is a good answer in the action's docs:
```
How the action behaves:

If there are changes (i.e. a diff exists with the checked-out base branch), the changes will be pushed to a new branch and a pull request created.
If there are no changes (i.e. no diff exists with the checked-out base branch), no pull request will be created and the action exits silently.
If a pull request already exists it will be updated if necessary. Local changes in the Actions workspace, or changes on the base branch, can cause an update. If no update is required the action exits silently.
If a pull request exists and new changes on the base branch make the pull request unnecessary (i.e. there is no longer a diff between the pull request branch and the base), the pull request is automatically closed. Additionally, if delete-branch is set to true the branch will be deleted.
```

from https://github.com/peter-evans/create-pull-request#action-behaviour 

TL;DR - **yes** - such scenarios are handled automagically by this action

.github/workflows/update-intellij-version.yml
LukaszKontowski: - if a branch already exists and no PR for this branch - then `peter-evans/create-pull-request@v4` will add another commit on existing branch and create a pull request
- if a branch exists and there is a PR for this branch - then `peter-evans/create-pull-request@v4` will add another commit on existing branch on the same PR

LukaszKontowski: `BRANCH_NAME` changed as suggested

.github/workflows/update-intellij-version.yml
PawelLipski: For my exp, `cut` is way more familiar to shell users than `awk` ;)

PawelLipski (approved):
Okay! should we expect a PR to open with the next cron run? Actually, I'd change the cron to more frequent (even daily) for a moment, so that we don't need to wait for that first run too long

LukaszKontowski:
> Okay! should we expect a PR to open with the next cron run? Actually, I'd change the cron to more frequent (even daily) for a moment, so that we don't need to wait for that first run too long

Yes, a PR should appear with the first cron run. For me could be daily or once a week, both sound good. :) @lukaszwawrzyk - what is your opinion on that? Should we run this job once a week or daily?

PawelLipski:
> Should we run this job once a week or daily?

I'd suggest running daily just for the first update, then reverting to weekly (daily is an overkill for updates that come out once in a few weeks/months at best)

LukaszKontowski:
Changed the schedule to daily as suggested

LukaszKontowski:
It might be possible to check compatibility of the Jetbrains Scala Plugin automatically - instead of adding the note in PR body telling that PR asignee should check the compatibility him/her self. It could be done by:

1. Listing git tags from the https://github.com/JetBrains/intellij-scala repo - choosing tags that have names corresponding to the new intellij release.
2. After few potentially valid git tags are chosen - for each of them:
- check the https://github.com/JetBrains/intellij-scala/blob/2022.3.1/pluginXml/resources/META-INF/plugin.xml (this one is for `2022.3.1` tag) - could be done by cloning git repo or HTML parsing
- take the `<idea-version since-build="????" until-build="????"/>` line from mentioned file
- compare `since-build` and `until-build` with new intelliJ build version number
- if matches - OK
- if does not match - check another potentially valid git tag and loop again

I just wonder if providing such logic is worth it. 🤔

LukaszKontowski:
For now abandoning idea of automated Scala Plugin version bumps. It would not be that helpful, as such changes would be necessary for **major** releases of IntelliJ IDEA only - so +- 3 times a year. And such solution would not be clean and stable probably.

PR #282, opened 2022-09-19
Make git clone logging less verbose

### LINKED ISSUE #274: sbt logs in CI are spammed with git pull output
See e.g. https://github.com/VirtusLab/ide-probe/runs/8303076011?check_suite_focus=true

image

Requires a looot of scrolling (which doesn't seem to work particularly smooth in GH actions)

LukaszKontowski:
Closes #274

These changes fix logging of `git-clone` related logs - so that there will be only one line logged for 1% of progress. Changes shorten the total log length for `tests (2.12, probe)` and `tests (2.13 probe)` from 24.000+ lines to less than 3.500 lines.

### DIFF
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/git/GitHandler.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/git/GitHandler.scala
@@ -5,6 +5,8 @@ import java.nio.file.Path
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.AtomicReference
 
+import scala.collection.mutable
+
 import org.eclipse.jgit.api.CloneCommand
 import org.eclipse.jgit.api.CreateBranchCommand
 import org.eclipse.jgit.api.Git
@@ -75,20 +77,20 @@ object GitHandler {
     private val completion: AtomicInteger = new AtomicInteger()
     private val task: AtomicReference[String] = new AtomicReference()
 
+    private val titlePercentArray: mutable.ArrayBuffer[(String, Int)] = mutable.ArrayBuffer.empty
+
     @inline
-    private def renderProgress(completed: Int): String = {
+    private def printProgress(completed: Int): Unit = {
       val total = totalWork.get()
       val title = task.get()
-      val inner =
-        if (total == 0)
-          ""
-        else if (completed == total)
-          s"$title 100% ($total/$total)\n"
-        else {
-          val percent = (100 * completed) / total
-          s"$title $percent% ($completed/$total)"
-        }
-      s"\r$inner"
+
+      if (total == 0) return
+
+      val percent = (100 * completed) / total
+      if (!titlePercentArray.contains((title, percent))) { // to print only one line per 1 percent in logs
+        titlePercentArray += (title -> percent)
+        println(s"$title $percent% ($completed/$total)")
+      }
     }
 
     override def start(totalTasks: Int): Unit = {
@@ -103,7 +105,7 @@ object GitHandler {
 
     override def update(completed: Int): Unit = {
       val current = completion.getAndUpdate(_ + completed)
-      print(renderProgress(current))
+      printProgress(current)
     }
 
     override def endTask(): Unit = {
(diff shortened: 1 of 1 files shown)


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/git/GitHandler.scala
```diff
@@ -75,20 +77,20 @@ object GitHandler {
     private val completion: AtomicInteger = new AtomicInteger()
     private val task: AtomicReference[String] = new AtomicReference()
 
+    private val titlePercentArray: mutable.ArrayBuffer[(String, Int)] = mutable.ArrayBuffer.empty
```
PawelLipski, line 80: Any risk of concurrent writes/race conditions here?

LukaszKontowski: I checked the logs and it doesn't look like our tests bring such risk.

PawelLipski (approved):

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/git/GitHandler.scala
```diff
@@ -75,20 +77,20 @@ object GitHandler {
...
-          ""
-        else if (completed == total)
-          s"$title 100% ($total/$total)\n"
-        else {
-          val percent = (100 * completed) / total
-          s"$title $percent% ($completed/$total)"
-        }
-      s"\r$inner"
+
+      if (total == 0) return
```
lukaszwawrzyk, line 87: We rather don't use return like that in scala

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/git/GitHandler.scala
```diff
@@ -75,20 +77,20 @@ object GitHandler {
...
+
+      val percent = (100 * completed) / total
+      if (!titlePercentArray.contains((title, percent))) { // to print only one line per 1 percent in logs
+        titlePercentArray += (title -> percent)
+        println(s"$title $percent% ($completed/$total)")
```
lukaszwawrzyk, line 92: what happened to print() and  \r?

PawelLipski: Huh `\r` was supposed to return to the beginning of the line, so that the previously written string could be overwritten right? 🤔 this doesn't work in most CIs, tho 🤔

lukaszwawrzyk: Yet it looks much better in terminal. The title says "Make git clone logging less verbose", while from perspective of terminal users it is more verbose as instead of replacing output it adds new for each %. Can we detect if output is connected to a tty (is it on ci?) or add some CI mode?

PawelLipski: +1, it probably can be just as easy as checking for `CI` env var

LukaszKontowski: Ok, @lukaszwawrzyk - I just fixed it.
- for local execution (terminal users) - logging will work exactly the same as before
- for github actions workflows - logging will use new logic (1 line per 1%)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/git/GitHandler.scala
```diff
@@ -83,13 +83,32 @@ object GitHandler {
     private def printProgress(completed: Int): Unit = {
       val total = totalWork.get()
       val title = task.get()
+      if (System.getenv("CI") == "true") // should always be "true" for github actions workflow
+        printForCI(title, completed, total)
+      else
+        printForNoCI(title, completed, total)
```
PawelLipski, line 89:
```suggestion
        printForNonCI(title, completed, total)
```

PawelLipski (approved):

lukaszwawrzyk (approved):

LukaszKontowski:
looks like some one-time error not related to PR changes https://github.com/VirtusLab/ide-probe/actions/runs/3242770504/jobs/5316526357

PR #283, opened 2022-09-20
Add support for installing IntelliJ from .dmg

### LINKED ISSUE #281: Add support for installing IDE from .dmg
Currently only installation from `.zip` is possible (despite the existence of `org.virtuslab.ideprobe.dependencies.IntelliJDmgResolver`)

LukaszKontowski:
Closes #281

Tested locally against `git-machete-intellij-plugin` - all tests pass properly with following config:
```
probe.resolvers {
    intellij.repositories = [
      "https://download.jetbrains.com/idea/ideaIC-2022.2.2.dmg"
    ]
  }
```

to be done:
- [x] add description for users in `reference.conf`

### DIFF
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -17,7 +17,7 @@ probe {
       // `ext` stands for the file extension (file format). This config specifies which extension should be
       // downloaded and installed. By default, ".zip" distributions of IntelliJ IDEA are downloaded. For now
       // two extensions are supported: ".zip" and ".dmg". ".zip" is the default one whereas ".dmg" shuld be used
-      // for osx only.
+      // for Mac OS X only.
       ext = ".zip"
       # ext = ".dmg"
 
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -16,8 +16,10 @@ probe {
 
       // `ext` stands for the file extension (file format). This config specifies which extension should be
       // downloaded and installed. By default, ".zip" distributions of IntelliJ IDEA are downloaded. For now
-      // only ".zip" extension is supported.
+      // two extensions are supported: ".zip" and ".dmg". ".zip" is the default one whereas ".dmg" shuld be used
+      // for osx only.
       ext = ".zip"
+      # ext = ".dmg"
 
     }
 
@@ -96,8 +98,8 @@ probe {
     // ]
     //
     // Moreover, you can add a pattern pointing to a directory of an installed IntelliJ instance instead of
-    // pattern pointing to a .zip file. ide-probe will recognize if the pattern points to a directory with
-    // installed IntelliJ or to a .zip file. For example, if you run the IntelliJ Plugin Verifier on your
+    // pattern pointing to a .zip (or .dmg) file. ide-probe will recognize if the pattern points to a directory with
+    // installed IntelliJ or to a .zip (or .dmg) file. For example, if you run the IntelliJ Plugin Verifier on your
     // repository (https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#tasks-runpluginverifier),
     // then you could use following configuration:
     // intellij.repositories = [
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -16,7 +16,7 @@ probe {
 
       // `ext` stands for the file extension (file format). This config specifies which extension should be
       // downloaded and installed. By default, ".zip" distributions of IntelliJ IDEA are downloaded. For now
-      // two extensions are supported: ".zip" and ".dmg". ".zip" is the default one whereas ".dmg" shuld be used
+      // two extensions are supported: ".zip" and ".dmg". ".zip" is the default one whereas ".dmg" should be used
       // for Mac OS X only.
       ext = ".zip"
       # ext = ".dmg"
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
@@ -92,8 +92,9 @@ object Resource extends ConfigFormat {
       import sys.process._
 
       // create tmp directory where disk image will be attached
-      val dmgDir = Paths.get(System.getProperty("java.io.tmpdir")).resolve("dmg_dir")
-      dmgDir.createDirectory()
+      val dmgDirPath = Paths.get(System.getProperty("java.io.tmpdir")).resolve("dmg_dir")
+      dmgDirPath.createDirectory()
+      val dmgDir = dmgDirPath.toString
 
       try {
         // attach disk image from .dmg file to local filesystem
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
@@ -91,17 +91,21 @@ object Resource extends ConfigFormat {
     def installTo(target: Path): Unit = {
       import sys.process._
 
-      // attach disk image from .dmg file to local filesystem
-      val dmgDir = s"${System.getProperty("java.io.tmpdir")}dmg_dir"
-      s"mkdir $dmgDir".!
-      s"hdiutil attach -mountpoint $dmgDir ${path.toString}".!
+      // create tmp directory where disk image will be attached
+      val dmgDir = Paths.get(System.getProperty("java.io.tmpdir")).resolve("dmg_dir")
+      dmgDir.createDirectory()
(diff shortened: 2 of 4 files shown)


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
```diff
@@ -86,6 +87,24 @@ object Resource extends ConfigFormat {
     }
   }
 
+  final class DMGFile(val path: Path) extends IntellijResource {
+    def installTo(target: Path): Unit = {
+      import sys.process._
+
+      // attach disk image from .dmg file to local filesystem
+      val dmgDir = s"${System.getProperty("java.io.tmpdir")}dmg_dir"
```
PawelLipski, line 95: Nit: is there a guarantee that `java.io.tmpdir` ends with a `/`? I'd rather trim the trailing path separator if present and then append it explicitly, or best use use path joining mechanism (like Python's `os.path.join`)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
```diff
@@ -86,6 +87,24 @@ object Resource extends ConfigFormat {
...
+      s"mkdir $dmgDir".!
+      s"hdiutil attach -mountpoint $dmgDir ${path.toString}".!
```
PawelLipski, line 97: Out of curiosity, what's the lifetime of such mount... until the machine is rebooted? (`/tmp/` might be special in this respect as its cleaned every now and then on reboot anyway; details differ between different Unix-like OSes, however)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
```diff
@@ -86,6 +87,24 @@ object Resource extends ConfigFormat {
...
+
+      // copy $dmgDir/IntelliJ IDEA CE.app to proper installation directory
+      val idePath = Paths.get(s"$dmgDir/IntelliJ IDEA CE.app/Contents")
+      idePath.copyDir(target)
+
+      // detach disk image from local filesystem
+      s"hdiutil detach $dmgDir".!
```
PawelLipski, line 104: Awww okay, nvm ;D

PawelLipski: Still, I'd use some mechanism like try-with-resources (or just a good old `try`-`finally`) to make sure that dmg is detached even if the copy fails for whatever reason

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
LukaszKontowski: Sure, switched to standard `resolve` method from `java.nio.file.Path` class

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
LukaszKontowski: I don't know that

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
LukaszKontowski: Ok, try-finally added

PawelLipski (approved):

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -16,8 +16,10 @@ probe {
 
       // `ext` stands for the file extension (file format). This config specifies which extension should be
       // downloaded and installed. By default, ".zip" distributions of IntelliJ IDEA are downloaded. For now
-      // only ".zip" extension is supported.
+      // two extensions are supported: ".zip" and ".dmg". ".zip" is the default one whereas ".dmg" shuld be used
+      // for osx only.
```
PawelLipski, line 20:
```suggestion
      // for Mac OS X only.
```

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -16,8 +16,10 @@ probe {
 
       // `ext` stands for the file extension (file format). This config specifies which extension should be
       // downloaded and installed. By default, ".zip" distributions of IntelliJ IDEA are downloaded. For now
-      // only ".zip" extension is supported.
+      // two extensions are supported: ".zip" and ".dmg". ".zip" is the default one whereas ".dmg" shuld be used
```
PawelLipski, line 19:
```suggestion
      // two extensions are supported: ".zip" and ".dmg". ".zip" is the default one whereas ".dmg" should be used
```

PawelLipski (approved):

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: misspells as always :P

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
```diff
@@ -86,6 +87,29 @@ object Resource extends ConfigFormat {
     }
   }
 
+  final class DMGFile(val path: Path) extends IntellijResource {
+    def installTo(target: Path): Unit = {
+      import sys.process._
+
+      // create tmp directory where disk image will be attached
+      val dmgDirPath = Paths.get(System.getProperty("java.io.tmpdir")).resolve("dmg_dir")
```
lukaszwawrzyk, line 95: can be written as 
      val dmgDirPath = Paths.get(System.getProperty("java.io.tmpdir")).createDirectory("dmg_dir")

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
```diff
@@ -86,6 +87,29 @@ object Resource extends ConfigFormat {
...
+      dmgDirPath.createDirectory()
+      val dmgDir = dmgDirPath.toString
+
+      try {
+        // attach disk image from .dmg file to local filesystem
+        s"hdiutil attach -mountpoint $dmgDir ${path.toString}".!
```
lukaszwawrzyk, line 101: for consistency, it would be better to use probe's shell api.
Shell.run("hdiutil", "attach", "-mountpoint", dmgDir.toString, path.toString)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/Resource.scala
```diff
@@ -86,6 +87,29 @@ object Resource extends ConfigFormat {
...
+        // copy $dmgDir/IntelliJ IDEA CE.app/ to proper installation directory
+        val idePath = Paths.get(s"$dmgDir/IntelliJ IDEA CE.app/Contents")
```
lukaszwawrzyk, line 103: It should also handle IntelliJ Ultimate. Maybe look for first dir starting with IntelliJ or something like that.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/ResourceProvider.scala
```diff
@@ -73,7 +74,11 @@ object ResourceProvider {
     }
 
     private def cached(uri: URI): Path = {
-      directory.resolve(Hash.md5(uri.toString))
+      val resultPath = directory.resolve(Hash.md5(uri.toString))
+      if (uri.toString.endsWith(".dmg"))
```
lukaszwawrzyk, line 78: you could make it more generic - to just keep extensions on these files. It would make checking this cache manually easier as well.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJProvider.scala
```diff
@@ -148,8 +148,14 @@ final case class IntelliJFactory(
     println(s"Installing $version")
     val file = dependencies.intelliJ.fetch(version)
     file.toExtracted.installTo(root)
-    root.resolve("bin").makeExecutableRecursively()
+    if (installedFromOsxDmg(root))
+      root.resolve("MacOS/idea").makeExecutable()
```
lukaszwawrzyk, line 152: Isn't MacOS dir present also in portable .zip distribution?

LukaszKontowski: You are right. It showed up that this change is not needed in fact, as `.makeExecutable()` on `MacOs/idea` is done inside `InstalledIntelliJ`

LukaszKontowski: I deleted this change from PR

lukaszwawrzyk (approved):

PR #285, opened 2022-09-26
Allow for globs in `probe.resolvers.intellij.repositories` config 

### LINKED ISSUE #250: Allow for globs in `probe.resolvers.intellij.repositories` config
Currently, in [git-machete-intellij-plugin -> ideprobe.conf](https://github.com/VirtusLab/git-machete-intellij-plugin/blob/master/src/uiTest/resources/ideprobe.conf#L18) we re-use the IntelliJs downloaded by `runPluginVerifier` Gradle task... but there's still no easy way to re-use IntelliJs downloaded by Gradle, which live under paths like:

```
$ ls -d ~/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/*/*/ideaIC-*/
/home/plipski/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/2022.2/42c296374014a649785bb84aa6d8dc2d18f2ca0e/ideaIC-2022.2/
/home/plipski/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/222.3345.16-EAP-SNAPSHOT/4a1d1d2eb2524a87cc7fd9bd7a2aed2c7efa6593/ideaIC-222.3345.16-EAP-SNAPSHOT/
/home/plipski/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/222.3345.47-EAP-SNAPSHOT/bf10a6ac5e737b257349ade298202da993b357ca/ideaIC-222.3345.47-EAP-SNAPSHOT/
/home/plipski/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/222.3345.90-EAP-SNAPSHOT/1edc98c7aa0e16fa9f95e2cd396a556b1a03aaf6/ideaIC-222.3345.90-EAP-SNAPSHOT/
/home/plipski/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/222.3739.24-EAP-SNAPSHOT/fd84e4e4a7796908228a8f72c6b156de1c5c916d/ideaIC-222.3739.24-EAP-SNAPSHOT/
```

Due to the hash in path, we can't easily define a single expression with `[revision]` to put into `probe.resolvers.intellij.repositories`.

Let's:
1. check if the Gradle-downloaded IntelliJ can indeed be used by ide-probe: just set `probe.resolvers.intellij.repositories` to a single element, a specific path from Gradle caches, for a specific version (the hash included)
2. if so, implement support for globs so that a path template like 
```
"file:///"${HOME}"/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/[revision]/*/ideaIC-[revision]/"
```
works for any version (snapshot or release)
3. update the docs accordingly

LukaszKontowski:
Closes #250
_work in progress_

to be done:
- [x] add description and some example for users in `reference.conf`
- [x] add one more test scenario to test new logic with multiple paths (but only one is valid)

### DIFF
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -109,6 +109,21 @@ probe {
     // ~/.pluginVerifier/ides/IC-[revision]/ directory (if exists). If it does not exist under specified path,
     // then ide-probe will download it from one of the 6 official repositories.
     //
+    // If your config contains a pattern pointing to an IntelliJ instance that exists on local filesystem, then you
+    // can use globs (* widcard characters) as parts of the pattern. One * wildcard character replaces
+    // one atomic directory on the path. Multiple * wildcards can be used in one pattern. For example - this can
+    // be useful if you want to use .zip OR installed IntelliJ downloaded earlier by some gradle task:
+    // intellij.repositories = [
+    //   "file:///"${HOME}"/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/[revision]/*/ideaIC-[revision]/"
+    // ]
+    // - as you can see - the * wildcard is handy as it replaces a hash that could not be known easily. There could be
+    // multiple different direcories in the * place - but not all of them lead to a valid IntelliJ resource. ide-probe
+    // deals with such situations and finds the correct path to a valid IntelliJ resource. Note: example below would
+    // work as well as the previous one, as multiple * wildcards in one path are handled properly:
+    // intellij.repositories = [
+    //   "file:///"${HOME}"/.gradle/*/modules-2/*/*/ideaIC/[revision]/*/ideaIC-[revision]/"
+    // ]
+    //
     // Note: if you use the `intellij.repositories` config with a value pointing to an installed
     // IntelliJ instance directory - then DO NOT use the `probe.intellij.path` config in the same time.
     // The `probe.intellij.path` config takes precedence over `intellij.repositories` in such case, so the path
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -118,8 +118,9 @@ probe {
     // ]
     // - as you can see - the * wildcard is handy as it replaces a hash that could not be known easily. There could be
     // multiple different directories in the * place - but not all of them lead to a valid IntelliJ resource. ide-probe
-    // deals with such situations and finds the correct path to a valid IntelliJ resource. Note: example below would
-    // work as well as the previous one, as multiple * wildcards in one path are handled properly:
+    // deals with such situations and finds the correct path to a valid IntelliJ resource. The pattern must start with
+    // "file:" if resolving * wildcards should be used.
+    // Note: example below would work as good as the previous one. Multiple * wildcards in one path are handled properly:
     // intellij.repositories = [
     //   "file:///"${HOME}"/.gradle/*/modules-2/*/*/ideaIC/[revision]/*/ideaIC-[revision]/"
     // ]
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -123,6 +123,7 @@ probe {
     // intellij.repositories = [
     //   "file:///"${HOME}"/.gradle/*/modules-2/*/*/ideaIC/[revision]/*/ideaIC-[revision]/"
     // ]
+    // Note: you can NOT use the * wildcard in place of the last element of the pattern (as that would lead to ambiguous results).
     //
     // Note: if you use the `intellij.repositories` config with a value pointing to an installed
     // IntelliJ instance directory - then DO NOT use the `probe.intellij.path` config in the same time.
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -117,7 +117,7 @@ probe {
     //   "file:///"${HOME}"/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/[revision]/*/ideaIC-[revision]/"
     // ]
     // - as you can see - the * wildcard is handy as it replaces a hash that could not be known easily. There could be
-    // multiple different direcories in the * place - but not all of them lead to a valid IntelliJ resource. ide-probe
+    // multiple different directories in the * place - but not all of them lead to a valid IntelliJ resource. ide-probe
     // deals with such situations and finds the correct path to a valid IntelliJ resource. Note: example below would
     // work as well as the previous one, as multiple * wildcards in one path are handled properly:
     // intellij.repositories = [
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
@@ -67,7 +67,7 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
     !candidatePath.contains('*') && // below - make sure that candidatePath's last path element is same as in pattern
       candidatePath.substring(candidatePath.lastIndexOf('/')) == originalPattern.substring(
         originalPattern.lastIndexOf('/')
-      )
+      ) && Paths.get(candidatePath.replace("file:", "")).exists
 
   private def replaceFirstFoundWildcardWithDirectories(path: String): List[String] = {
     def removeLastFileSeparator(path: String): String = if (path.endsWith("/")) path.init else path
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
@@ -1,7 +1,10 @@
 package org.virtuslab.ideprobe.dependencies
 
+import java.nio.file.Paths
+
 import org.virtuslab.ideprobe.IntelliJFixture
 import org.virtuslab.ideprobe.config.DependenciesConfig
+import org.virtuslab.ideprobe.Extensions.PathExtension
 
 trait IntelliJResolver {
   def community: DependencyResolver[IntelliJVersion]
@@ -37,9 +40,37 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
       "version" -> version.releaseOrBuild,
       "release" -> version.release.getOrElse("snapshot-release")
     )
-    val replaced = replacements.foldLeft(pattern) { case (path, (pattern, replacement)) =>
+    val replacedBeforeResolvingGlobs = replacements.foldLeft(pattern) { case (path, (pattern, replacement)) =>
       path.replace(s"[$pattern]", replacement)
     }
+    val replaced = replaceGlobsWithExistingDirectories(List(replacedBeforeResolvingGlobs)).head
     Dependency(replaced)
+  }
+
+  // solution below assumes that each * character is used to mark one part of the path (one atomic directory),
+  // for example: "file:///home/.cache/ides/com.jetbrains.intellij.idea/ideaIC/[revision]/*/ideaIC-[revision]/"
+  private def replaceGlobsWithExistingDirectories(paths: List[String]): List[String] =
+    if (paths.exists(pathMightBeValidResource))
+      paths.filter(pathMightBeValidResource)
+    else
+      paths.flatMap(replaceFirstFoundWildcardWithDirectories)
+
+  private def pathMightBeValidResource(path: String): Boolean =
+    path.indexOf('*') == -1 &&
+      (path.endsWith(".zip") ||
+        path.endsWith(".dmg") ||
+        Paths.get(path.replace("file:", "")).isDirectory)
+
+  private def replaceFirstFoundWildcardWithDirectories(path: String): List[String] = {
+    def removeLastFileSeparator(path: String): String = if (path.endsWith("/")) path.init else path
+
+    val pathUntilWildcard = path.substring(0, path.indexOf('*'))
+    val pathAfterWildcard = path.substring(path.indexOf('*') + 1)
+    Paths
+      .get(pathUntilWildcard.replace("file:", ""))
+      .directChildren()
+      .map { replaced =>
+        removeLastFileSeparator(replaced.toString) + pathAfterWildcard
+      }
   }
 }
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
@@ -78,8 +78,8 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
   private def replaceFirstFoundWildcardWithDirectories(path: String): List[String] = {
     def removeLastFileSeparator(path: String): String = if (path.endsWith("/")) path.init else path
 
-    val Array(stringPathUntilWildcard, stringPathAfterWildcard) = path.split("*", 2)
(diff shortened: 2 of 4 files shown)


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
```diff
@@ -37,9 +43,44 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
...
+  private def pathMightBeValidResource(path: String): Boolean =
+    path.indexOf('*') == -1 &&
+      (path.endsWith(".zip") ||
+        path.endsWith(".dmg") ||
+        Paths.get(path.replace("file:", "")).isDirectory)
+
+  private def replaceFirstFoundWildcardWithDirectories(path: String): List[String] = {
+    def removeLastFileSeparator(path: String): String = if (path.endsWith("/")) path.init else path
+
+    val pathUntilWildcard = Paths.get(path.substring(0, path.indexOf('*')).replace("file:", ""))
```
PawelLipski, line 71: Okay... so a ready library for globs wasn't an adequate solution here? 🤔

LukaszKontowski: I did not manage to test these libraries yet. Wanted to have a working solution first. I'm not sure if we need them here. Suggested docs https://docs.oracle.com/javase/tutorial/essential/io/find.html don't seem helpful to me in this case.

PawelLipski: Ok, leaving up to your judgement

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -109,6 +109,21 @@ probe {
...
     //
+    // If your config contains a pattern pointing to an IntelliJ instance that exists on local filesystem, then you
+    // can use globs (* widcard characters) as parts of the pattern. One * wildcard character replaces
+    // one atomic directory on the path. Multiple * wildcards can be used in one pattern. For example - this can
+    // be useful if you want to use .zip OR installed IntelliJ downloaded earlier by some gradle task:
+    // intellij.repositories = [
+    //   "file:///"${HOME}"/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/[revision]/*/ideaIC-[revision]/"
+    // ]
+    // - as you can see - the * wildcard is handy as it replaces a hash that could not be known easily. There could be
+    // multiple different direcories in the * place - but not all of them lead to a valid IntelliJ resource. ide-probe
```
PawelLipski, line 120:
```suggestion
    // multiple different directories in the * place - but not all of them lead to a valid IntelliJ resource. ide-probe
```

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -109,6 +109,21 @@ probe {
     // ~/.pluginVerifier/ides/IC-[revision]/ directory (if exists). If it does not exist under specified path,
     // then ide-probe will download it from one of the 6 official repositories.
     //
+    // If your config contains a pattern pointing to an IntelliJ instance that exists on local filesystem, then you
```
PawelLipski, line 112: :benissimo: the doc

PawelLipski (approved):

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
```diff
@@ -37,9 +43,44 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
       "version" -> version.releaseOrBuild,
       "release" -> version.release.getOrElse("snapshot-release")
     )
-    val replaced = replacements.foldLeft(pattern) { case (path, (pattern, replacement)) =>
+    val replacedBeforeResolvingGlobs = replacements.foldLeft(pattern) { case (path, (pattern, replacement)) =>
       path.replace(s"[$pattern]", replacement)
     }
+    val replaced = replaceGlobsWithExistingDirectories(List(replacedBeforeResolvingGlobs)).head
```
lukaszwawrzyk, line 49: shouldn't there be a function that accepts one argument and returns one value

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
```diff
@@ -37,9 +43,44 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
...
+  @tailrec
+  private def replaceGlobsWithExistingDirectories(paths: List[String]): List[String] =
+    if (paths.exists(pathMightBeValidResource))
+      paths.filter(pathMightBeValidResource)
+    else
+      replaceGlobsWithExistingDirectories(paths.flatMap(replaceFirstFoundWildcardWithDirectories))
+
+  private def pathMightBeValidResource(path: String): Boolean =
+    path.indexOf('*') == -1 &&
+      (path.endsWith(".zip") ||
```
lukaszwawrzyk, line 64: isn't it too specific check? Filling extension is responsibility of pattern replacement that already happened.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
```diff
@@ -37,9 +43,44 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
...
+  // for example: "file:///home/.cache/ides/com.jetbrains.intellij.idea/ideaIC/[revision]/*/ideaIC-[revision]/"
+  @tailrec
+  private def replaceGlobsWithExistingDirectories(paths: List[String]): List[String] =
+    if (paths.exists(pathMightBeValidResource))
+      paths.filter(pathMightBeValidResource)
+    else
+      replaceGlobsWithExistingDirectories(paths.flatMap(replaceFirstFoundWildcardWithDirectories))
+
+  private def pathMightBeValidResource(path: String): Boolean =
+    path.indexOf('*') == -1 &&
```
lukaszwawrzyk, line 63: path.contains('*')

PawelLipski: Negated, probably ;)

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
```diff
@@ -37,9 +43,44 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
...
+      (path.endsWith(".zip") ||
+        path.endsWith(".dmg") ||
+        Paths.get(path.replace("file:", "")).isDirectory)
```
lukaszwawrzyk, line 66: why don't you construct path from URI properly instead of manual conversion. Also you are leaving trailing slashes now.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
```diff
@@ -37,9 +43,44 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
...
+
+  private def replaceFirstFoundWildcardWithDirectories(path: String): List[String] = {
+    def removeLastFileSeparator(path: String): String = if (path.endsWith("/")) path.init else path
+
+    val pathUntilWildcard = Paths.get(path.substring(0, path.indexOf('*')).replace("file:", ""))
+    val stringPathAfterWildcard = path.substring(path.indexOf('*') + 1)
```
lukaszwawrzyk, line 72: can you use split for that? It allows to specify limit of splits.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
LukaszKontowski: Yes, the check is needed to find proper path from multiple directories that exist in place of `*` (some are not valid).
But I just realised that this check is not valid. This check should return `true` only when `path` ends properly, as defined in the pattern... Let me fix this.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
LukaszKontowski: Hm, I don't know if it would be possible. `replaceGlobsWithExistingDirectories` is recursive and returns multiple paths as a result - by replacing wildcards.

LukaszKontowski:
Ok, looks like logic is not 100% valid. I will undraft it after fix.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
LukaszKontowski: One thing I can do is to wrap `replaceGlobsWithExistingDirectories` in another helper function that accepts one argument and returns one value. But the  `replaceGlobsWithExistingDirectories`'s argument- and return- types must stay the same.

lukaszwawrzyk: Yes, please do. It is also quite common practice for such helper recursive function to be nested inside the actual function with expected api (to hide implementation details of recurrence).

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
LukaszKontowski: Ok, but then I would have to use three `val`s instead of two:
```
val Array(stringPathUntilWildcard, stringPathAfterWildcard) = path.split('*')
val pathUntilWildcard = Paths.get(stringPathUntilWildcard.replace("file:", ""))
```
If you think that would be better than current solution, I can do that

lukaszwawrzyk: Yes, let's do this. It is same number of lines, but 2 substring calls with some index offset are more complex for reading and understanding than a simple split.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
LukaszKontowski: Ok, logic of `pathMightBeValidResource` method is fixed now.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
LukaszKontowski: There is a problem connected to replacing `String#indexOf` with `String#split`
The `String#split(String regex, int limit)` method treats the String parameter `"*"` as a regular expression, which leads to invalid results. 

I would suggest to switch back to previous logic:
```
val pathUntilWildcard = Paths.get(path.substring(0, path.indexOf('*')).replace("file:", ""))
val stringPathAfterWildcard = path.substring(path.indexOf('*') + 1)
```
to be sure that `*` wildcard will not be used as a regular expression.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
```diff
@@ -37,9 +43,54 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
...
+
     Dependency(replaced)
   }
+
+  private def resolveGlobsInPattern(pathPattern: String): String =
+    replaceGlobsWithExistingDirectories(List(pathPattern), pathPattern).head
+
+  // Solution below assumes that each * character is used to mark one part of the path (one atomic directory),
+  // for example: "file:///home/.cache/ides/com.jetbrains.intellij.idea/ideaIC/[revision]/*/ideaIC-[revision]/".
+  // Wildcard character should NOT be used as the last element of the pattern - to avoid ambiguous results.
```
PawelLipski, line 63: Hmmm technically, ambiguous results can also happen when wildcard is somewhere inside the path as well, right?

Maybe sth like:
```suggestion
  // Wildcard character should NOT be used as the last element of the pattern - to decrease the probability of getting ambiguous results.
```

LukaszKontowski: > ambiguous results can also happen when wildcard is somewhere inside the path as well, right?

No, if the `*` wildcard is used in replacement of an atomic directory from the path (but not the last part) - then 
```
private def resolveGlobsInPattern(pathPattern: String): String =
    replaceGlobsWithExistingDirectories(List(pathPattern), pathPattern).head
```
- will return only one element if pattern is valid (so the result will not be ambiguous)
or
- will throw `NoSuchElementException` if pattern is not valid

PawelLipski: Huh okay... so let's break down all cases:

1. zero glob matches: exception regardless of where the glob is located?
2. exactly one glob match: OK
3. more than one glob match:
  a. if `*` is used for the last part of the glob &mdash; multiple results returned?
  b. if `*` is used for other part of the glob &mdash; exception thrown?

LukaszKontowski: Answered in PMs but short answers below:

> 1. zero glob matches: exception regardless of where the glob is located?

`NoSuchElementException` will be thrown (as pattern is not valid - such file does not exist)

> 2. exactly one glob match: OK

yes

> 3. more than one glob match:
a. if * is used for the last part of the glob — multiple results returned?

**no** - `NoSuchElementException` will be thrown (as pattern is not valid - such file does not exist) - using `*` as the last element of the path is not possible, because then the logic would not be able to choose the proper path.
For example - if we would have to path endings like below:
```
.../df364b7662aeb20ca2eafead730380017c395661/ideaIC-223.4884.69-EAP-SNAPSHOT 
.../df364b7662aeb20ca2eafead730380017c395661/ideaIC-223.4884.69-EAP-SNAPSHOT.zip
.../df364b7662aeb20ca2eafead730380017c395661/ideaIC-223.4884.69-EAP-SNAPSHOT.jar
.../df364b7662aeb20ca2eafead730380017c395661/ideaIC-223.4884.69-EAP-SNAPSHOT.pom
```
where first one is directory and second one - a `.zip` file - and the pattern ends with `*` - how can we know, which one should be used?

> 3. more than one glob match:
b. if * is used for other part of the glob — exception thrown?

**no** - `resolveGlobsInPattern` will choose the first one by calling `.head`

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
lukaszwawrzyk: To be sure, just use regex as expected r"\*" or "[*]"

lukaszwawrzyk (approved):

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
LukaszKontowski: Ok, but the thing is - I don't want java String's method to treat the `*` as a regular expression. And `String#split` by default treats the String parameter as a regular expression. That's why I want to stick to current logic, without using `split`. So that `*` will never be treated as a regex.

PR #300, opened 2022-10-25
Add full documentation to the `reference.conf` file

### LINKED ISSUE #254: Move docs on HOCON config from docs/ to comments in reference.conf
That's where it'd be more expected to be found, from what I've seen across the libs in Scala ecosystem

### LINKED ISSUE #262: Add proper descriptions and examples for all values in `reference.conf`

### LINKED ISSUE #265: If both `probe.intellij.version.build` and `.release` are specified, which one takes precedence?
In particular, what happens if they point to different builds?

Let's then think what can be simplified in the config... in particular, it'd be very useful it one of this keys could be explicitly used as "build or release", i.e. accept both formats. In fact, `build` can be used in this way currently (at least when `release` is NOT defined)... let's maybe make this behavior documented, if it can be relied upon

LukaszKontowski:
Closes #254 
Closes #262 
Closes #265

Adds full documentation to the `reference.conf`.
4 files from `docs/` directory are deleted as their content (more or less modified) is now available in the `reference.conf` config comments.

One thing still missing is a description with example for usage of
`org.virtuslab.ideprobe.dependencies.Plugin.FromSources` which could be used in `probe.intellij.plugins` array. The reason is that for now this one is not clear to me and I did not manage to find a single example of usage for this type of `Plugin`. Moreover, it looks like `FromSources` usage is dedicated to users, who interact with ide-probe rather by direct code calls (and not by the `.conf` file), because it requires the user to have a custom implementation of `DependencyBuilder` and register it by using `org.virtuslab.ideprobe.dependencies.DependencyProvider#registerBuilder`...

If you have some suggestions / advice about it, that would be nice. Otherwise - I can create a separate issue for adding a proper comment for this one missing element.


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJVersion.scala
```diff
@@ -6,8 +6,6 @@ import pureconfig.generic.semiauto.deriveConvert
 import org.virtuslab.ideprobe.IntelliJFixture
 import org.virtuslab.ideprobe.config.IntellijConfig
 
-// the `ext` field (file extension) must be an Option[String] and not a String for the ExistingIntelliJ usage
-// inside the `org.virtuslab.ideprobe.dependencies.IntelliJVersionResolver` object
```
LukaszKontowski, line 10: unrelated - deleted outdated invalid code comment

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -5,12 +5,52 @@ probe {
 
   // `intellij` is for org.virtuslab.ideprobe.config.IntellijConfig - default values below:
   intellij {
+    // `plugins` array contains IntelliJ plugins which should be installed and enabled on IntelliJ run. Here you can add
+    // your IntelliJ plugin for tests (amongst other plugins). There are a few ways to provide a proper entry pointing
+    // to a plugin:
+    // 1. By using a direct URI pointing to the plugin (could be from a website or from a local file). For example:
+    #     { uri = "https://packages.jetbrains.team/maven/p/ij/intellij-dependencies/com/intellij/remoterobot/robot-server-plugin/0.11.15/robot-server-plugin/0.11.15.zip" }
+    // 2. By using default IntelliJ plugins repository (see `probe.resolvers.plugins.repository.uri`) with additional
```
PawelLipski, line 13: If `probe.resolvers.plugins.repository.uri` is **changed from the default**, then the is _default IntelliJ plugins repository_ or the newly overridden repository used for resolving the plugins? (I suppose the latter, but the use of _default_ in this sentence suggests the former)

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -5,12 +5,52 @@ probe {
...
+    //    id (pluginId), version and (optional) channel parameters which is translated into a URL with a query string.
+    //    For example:
+    #       { id = "org.intellij.scala", version = "2020.3.369", channel = "nightly" }
+    //      will be resolved into following URL with query string (from where ide-probe should download the plugin):
+    //      https://plugins.jetbrains.com/plugin/download?pluginId=org.intellij.scala&version=2020.3.369&channel=nightly
+    //    and without the optional "channel" parameter:
+    #       { id = "org.intellij.scala", version = "2020.3.369" }
+    //      will be resolved into following URL with query string (from where ide-probe should download the plugin):
+    //      https://plugins.jetbrains.com/plugin/download?pluginId=org.intellij.scala&version=2020.3.369
+    // 3. By using a bundled IntelliJ plugin. Such plugin should be bundled in your project as a module/subproject and
```
PawelLipski, line 23: Huh _bundled_ already has a specific meaning for IntelliJ plugins: see Plugins in Settings:

image

I wouldn't use this word to avoid confusion

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -5,12 +5,52 @@ probe {
...
+    //    available on the classpath when IntelliJ instance should run. For example - such plugin subproject should be
+    //    built and published to a `target` directory before performing tests (during compilation).
+    //    There are two ways to point to a bundled IntelliJ plugin:
+    //    a) By providing plugin's resource file name (probably plugin subproject's name followed by version and file extension).
+    //       For example:
+    #        { bundle = "driver-test-plugin_2.13-0.45.0-SNAPSHOT.zip" }
```
PawelLipski, line 29: And the use of `bundle` in config also isn't fortunate... but it's a different story 😅

PawelLipski: Huh unless `bundled` here and `bundled` as in, _included with IntelliJ by default_ somehow correspond to the same category of plugins? 🤔 but it isn't obvious

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -5,12 +5,52 @@ probe {
...
+    //       whch will be resolved properly if mentioned .zip file is available on the classpath.
+    //    b) By providing three values realted to the plugin: name, scalaVersion (major) and version. For example:
```
PawelLipski, line 31:
```suggestion
    //    b) By providing three plugin coordinates: name, scalaVersion (major) and version. For example:
```

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -5,12 +5,52 @@ probe {
...
     plugins = []
+
+    // `version.release` is the "Version" and connected `version.build` is the "Build Number" of IntelliJ IDEA version.
+    // You can search for proper string pairs here:
+    // https://www.jetbrains.com/intellij-repository/releases/ under "com.jetbrains.intellij.idea" heading.
+    // Moreover - builds from https://www.jetbrains.com/intellij-repository/snapshots/ can be used by default as well.
+    // If you want to change the default intellij version, provide both parameters in your .conf file.
+    // Notes:
+    // - `version.release` takes precedence over `version.build`. If both values point to two different builds, then
+    //    `version.release`- related build will be used.
```
PawelLipski, line 43:
```suggestion
    //    the build corresponding to `version.release` will be used.
```

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -5,12 +5,52 @@ probe {
...
+    //    b) By providing three values realted to the plugin: name, scalaVersion (major) and version. For example:
+    #        { bundle = "driver-test-plugin", scalaVersion = "2.13", version = "0.45.0-SNAPSHOT" }
+    //       which will be resolved the same as the earlier example (so, `.zip` extension is th default one).
     plugins = []
+
+    // `version.release` is the "Version" and connected `version.build` is the "Build Number" of IntelliJ IDEA version.
+    // You can search for proper string pairs here:
+    // https://www.jetbrains.com/intellij-repository/releases/ under "com.jetbrains.intellij.idea" heading.
+    // Moreover - builds from https://www.jetbrains.com/intellij-repository/snapshots/ can be used by default as well.
+    // If you want to change the default intellij version, provide both parameters in your .conf file.
```
PawelLipski, line 40: Nit: use `IntelliJ` consistently

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -161,34 +201,108 @@ probe {
       "https://cache-redirector.jetbrains.com/intellij-jbr/jbr-[major]-[platform]-x64-b[minor].tar.gz"
     ]
 
+    // The maximal number of retries that should be performed by resolvers when resolving a pattern to a resource.
```
PawelLipski, line 204: The retries are only relevant for **network requests** right?

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -161,34 +201,108 @@ probe {
...
 
   // `driver` is for `org.virtuslab.ideprobe.config.DriverConfig` - default values below
   driver {
+
+    // `driver.launch` contains settings that define how IDE should be started.
     launch {
+
+      // `driver.launch.command` is a string array transformed into the command used to launch the IDE (each string
+      // from the array is a single element of the command, spaces are inserted between the elements). By default,
+      // the driver uses the `idea` command, but you can specify a custom command by using this config.
```
PawelLipski, line 216: Should it then be `command = ["idea"]`?

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -161,34 +201,108 @@ probe {
...
+        // should be ignored even if `driver.check.errors.enabled` is set to `true`.
+        //
+        // Both for `includeMessages` and for `excludeMessages` you can define an entry in one of two possible ways:
+        // 1. By using a regex-based string. For example: ".*at com.intellij.uast.UastMetaLanguage.*" will match
+        //    any error message that contains "at com.intellij.uast.UastMetaLanguage". Note: your regex-based string
+        //    CAN NOT be a multi-line string.
+        // 2. By using a raw string without regular expressions. In such case ide-probe just checks if IDE error message
+        //    contains your raw string. The benefit of this option is that you can use a multi-line string (part of
+        //    stacktrace). You don't have to worry about line indentations here as each line is trimmed before performing
+        //    the check. If you want to use a multi-line stirng, use triple quotes: """<your multi-line string...>""".
```
PawelLipski, line 255:
```suggestion
        //    the check. If you want to use a multi-line string, use triple quotes: """<your multi-line string...>""".
```

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -161,34 +201,108 @@ probe {
...
       }
+
+      // `driver.check.freezes.enabled` defines if ide-probe should throw an exception and fail the test if IDE freezes
+      // have occured during test. By default, this functionality is disabled. Set to `true` to enable.
       freezes.enabled = false
     }
+
+    // `driver.display` defines the display mode which should be used for the IDE run. By default, it is "native",
+    // which means that everything is displayed on the screen and you can interact with the IDE during test.
+    // Another possible value is "xvfb" which makes the IDE run in the xvfb mode (no display). This config can be set
```
PawelLipski, line 270:
```suggestion
    // Another possible value is "xvfb" which makes the IDE run in the xvfb mode (using virtual display; only available under X Window System). This config can be set
```

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -161,34 +201,108 @@ probe {
...
+    // either with a raw string value or by setting the IDEPROBE_DISPLAY environment variable which overrides the default value.
     display = "native" # OR "xvfb" - these are two possible values
     display = ${?IDEPROBE_DISPLAY}
+
+    // `driver.xvfb` contains parameters that are used if IDE runs in the xvfb mode. Basically, it contains three
+    // screen size parameters that are passed to the "−screen" command line switch of xvfb.
```
PawelLipski, line 276: Strictly speaking, they're passed to [`xvfb-run --server-args="..."`](https://manpages.ubuntu.com/manpages/xenial/man1/xvfb-run.1.html), which then subsequently passes them to `Xvfb`

PawelLipski: Coz `xvfb-run` runs two processes:
1. Xvfb (virtual display server)
2. the specified process, with X display config pointing to (1.)

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -161,34 +201,108 @@ probe {
...
+    // `driver.headless` defines if IDE should run in a headless mode (without a graphical user interface). If you want
+    // to use it, set to `true`. Specifically, it adds " headless" to the basic launch command ("idea").
+    // Note: if `headless` is set to `true`, it takes precedence over `driver.display`, which means that
+    // IDE will run in the headless mode (and not in a "native" or "xvfb" mode).
+    // Additionally, unless run in headless mode, the driver captures the screenshots of the screen periodically,
+    // so it can then be exposed on the CI for the user to inspect the IDE state in case of failures. If you run in the
+    // headless mode - then no screenshots will be captured.
     headless = false
+
+    // `driver.vmOptions` are java virtual machine (JVM) options that are used by the driver to further customize the IDE.
```
PawelLipski, line 295:
```suggestion
    // `driver.vmOptions` are Java Virtual Machine (JVM) options that are used by the driver to further customize the IDE.
```

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: The value from `probe.resolvers.plugins.repository.uri` will be used as the _default IntelliJ plugins repository_ to resolve the Plugin to a valid URL. Let me make it more clear.

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: Ok, but this is ide-probe's naming convention. These "ways" of resolving Plugins are in fact called "Bundled" in code. What do you suggest?
image

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: But now I see that it leads to confusion. Maybe a short introduction to ide-probe's _bundle_ naming convention will be needed here?

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: 👍 In my opinion - yes. We can create an issue for that need. In fact, current solution allows both versions and treats them the same. But the `command` was an empty array before creating `reference.conf`, that's why it is what it is now. However, I think we can do a fast cleanup and set `command = ["idea"]` as the default one (with some small changes).
image

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -161,34 +201,108 @@ probe {
...
     xvfb {
       screen {
         width = 1920
         height = 1080
         depth = 24
       }
     }
+
+    // `driver.headless` defines if IDE should run in a headless mode (without a graphical user interface). If you want
+    // to use it, set to `true`. Specifically, it adds " headless" to the basic launch command ("idea").
```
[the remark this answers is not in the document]
LukaszKontowski: It looks so in the code.
image

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -161,34 +201,108 @@ probe {
...
+    // Note: if `headless` is set to `true`, it takes precedence over `driver.display`, which means that
```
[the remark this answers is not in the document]
LukaszKontowski: Yes, these are two different ways of running IDE. In fact, it seems to me that we could fix the way ide-probe is configured so that you can choose in only one config, which way should be chosen:
a) native
b) xvfb
c) headless mode
image

core/driver/sources/src/main/resources/reference.conf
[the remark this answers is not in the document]
LukaszKontowski: Also, tested it locally - such usage is possible for IntelliJ
image

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: In the standard ide-probe processing logic - yes. However, it is possible that `retries` will be used for a non-network request resolving. It is also used (for some reason, which I do not understand) for **jar** URIs.
image

LukaszKontowski: But I'm not sure if a retry would make sense for such **jar** scenario to be honest. @lukaszwawrzyk - can you help us understand this part?

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -161,34 +201,108 @@ probe {
...
   }
 
   // `driver` is for `org.virtuslab.ideprobe.config.DriverConfig` - default values below
   driver {
+
+    // `driver.launch` contains settings that define how IDE should be started.
     launch {
+
+      // `driver.launch.command` is a string array transformed into the command used to launch the IDE (each string
+      // from the array is a single element of the command, spaces are inserted between the elements). By default,
```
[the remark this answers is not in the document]
LukaszKontowski: The process and its launch command is in the end prepared by `com.zaxxer.nuprocess.NuProcessBuilder`. In fact we pass `driver.launch.command` array (raw or with first element - "idea" command - expanded with `xvfb` / `headless` additions) to `NuProcessBuilder` constructor so that `NuProcessBuilder`'s field named `command` is equal to what we pass there.
And here is a javadoc about this field and how it is used:
```
a command, a list of strings which signifies the external program file to be invoked and its arguments, if any. Which string lists represent a valid operating system command is system-dependent. For example, it is common for each conceptual argument to be an element in this list, but there are operating systems where programs are expected to tokenize command line strings themselves - on such a system a Java implementation might require commands to contain exactly two elements.
```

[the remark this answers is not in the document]
LukaszKontowski: To be honest, I don't know the answer.

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: Ok, but if I get it right, `xvfb-run` is a _wrapper_ for the `Xvfb` (1x - for now it's version 11). And in my comments I added a link to the https://www.x.org/releases/X11R7.6/doc/man/man1/Xvfb.1.xhtml docs. And in this manual, they say:
```
In addition to the normal server options described in the [Xserver(1)](https://www.x.org/releases/X11R7.6/doc/man/man1/Xserver.1.xhtml)
manual page, Xvfb accepts the following command line switches:
−screen screennum WxHxD

This option creates screen screennum and sets its width, height, and depth to W, H, and D respectively. By default, only screen 0 exists and has the dimensions 1280x1024x8.
...
```
So, I think we can say that these 3 params are passed to the "−screen" command line switch of Xvfb?

LukaszKontowski: BTW, just changed from "xvfb" to "Xvfb" on this line to be more precise

core/driver/sources/src/main/resources/reference.conf
PawelLipski: Pls open an issue to rethink this naming. No action to be done on this PR yet

core/driver/sources/src/main/resources/reference.conf
PawelLipski: Yeah, pls open an issue for the ☝🏻

core/driver/sources/src/main/resources/reference.conf
PawelLipski: Huh kinda makes sense for the rare case of network file systems (so a `file:///`-like URI, which still requires an access over network)... maybe doesn't harm to leave as is indeed

core/driver/sources/src/main/resources/reference.conf
PawelLipski: Ok, +1

core/driver/sources/src/main/resources/reference.conf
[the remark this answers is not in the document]
PawelLipski: Wow okay... and the `launcher` point to some JetBrains provided script/executable? or is it sth generated by ide-probe itself?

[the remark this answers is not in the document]
PawelLipski: Okay, I see in `org.virtuslab.ideprobe.ide.intellij.InstalledIntelliJ#executable`, that's JetBrains-provided. Let's leave as is then

core/driver/sources/src/main/resources/reference.conf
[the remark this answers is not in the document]
PawelLipski: +1, makes sense

core/driver/sources/src/main/resources/reference.conf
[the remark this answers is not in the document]
LukaszKontowski: It points to the same script/executable. My previous screenshot was an example of a manual command execution in terminal. I did not use ide-probe, just did run `intellij headless` on my IntelliJ IDEA instance.

PawelLipski (approved):
:benissimo:

LukaszKontowski:
> Aren't there references to file that you deleted? Perhaps now readme show direct users to this reference.conf file?

Thanks for suggestion - I will check that if there references to files that I deleted before merging. If they do exist - I will fix README in a separate PR soon.

PR #315, opened 2022-10-31
Move headless mode config from `driver.headless` to `driver.display`

### LINKED ISSUE #306: Change the way display mode is configured
Below are three possible ways how ide-probe can be run (regarding display mode)
a) `native`
b) `xvfb`
c) `headless` mode
But, for some reason, we have two different configs for this part of logic:
`probe.driver.display` to use `native` or `xvfb`
and
`probe.driver.headless` to decide if `headless` mode should be used 
And if `probe.driver.headless == true` then values from `probe.driver.display` are in fact never used.
image

It is confusing for users, who want to configure ide-probe with a `.conf` file. We can change the way `ide-probe`'s display mode is configured so that it is set in **only one config** (for example - get rid of the `probe.driver.headless` config and add support for "headless" value in the `probe.driver.display` config).

LukaszKontowski:
Closes #306

### DIFF
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -286,14 +286,16 @@ probe {
     // - By default, it is "native", which means that everything is displayed on the screen and you can interact
     //   with the IDE during test.
     // - Second possible value is "xvfb" which makes the IDE run in the xvfb mode (using virtual display; only available
-    //   under X Window System). This config can be set either with a raw string value or by setting the IDEPROBE_DISPLAY
-    //   environment variable which overrides the default value.
+    //   under X Window System).
     // - Third possible value is "headless" which runs IDE in a headless mode (without a graphical user interface).
     //   It adds " headless" option to the basic launch command ("idea").
     //
-    // Note: unless run in headless mode, the driver captures the screenshots of the screen periodically,
-    // so it can then be exposed on the CI for the user to inspect the IDE state in case of failures. If you run in the
-    // headless mode - then no screenshots will be captured.
+    // This config can be set either with a raw string value or by setting the IDEPROBE_DISPLAY
+    // environment variable which overrides the default value.
+    //
+    // Note: unless run in headless mode, the driver captures screenshots of the screen periodically,
+    // so it can then be exposed on the CI for the user to inspect the IDE state in case of failures.
+    // If you run in the headless mode - then no screenshots will be captured.
     display = "native" # OR "xvfb" OR "headless" - these are three possible values
     display = ${?IDEPROBE_DISPLAY}
 
--- a/core/driver/sources/src/main/resources/reference.conf
+++ b/core/driver/sources/src/main/resources/reference.conf
@@ -282,12 +282,19 @@ probe {
       freezes.enabled = false
     }
 
-    // `driver.display` defines the display mode which should be used for the IDE run. By default, it is "native",
-    // which means that everything is displayed on the screen and you can interact with the IDE during test.
-    // Another possible value is "xvfb" which makes the IDE run in the xvfb mode (using virtual display; only available
-    // under X Window System). This config can be set either with a raw string value or by setting the IDEPROBE_DISPLAY
-    // environment variable which overrides the default value.
-    display = "native" # OR "xvfb" - these are two possible values
+    // `driver.display` defines the display mode which should be used for the IDE run.
+    // - By default, it is "native", which means that everything is displayed on the screen and you can interact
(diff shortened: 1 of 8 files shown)


core/driver/sources/src/main/resources/reference.conf
```diff
@@ -282,12 +282,19 @@ probe {
...
-    // which means that everything is displayed on the screen and you can interact with the IDE during test.
-    // Another possible value is "xvfb" which makes the IDE run in the xvfb mode (using virtual display; only available
-    // under X Window System). This config can be set either with a raw string value or by setting the IDEPROBE_DISPLAY
-    // environment variable which overrides the default value.
-    display = "native" # OR "xvfb" - these are two possible values
+    // `driver.display` defines the display mode which should be used for the IDE run.
+    // - By default, it is "native", which means that everything is displayed on the screen and you can interact
+    //   with the IDE during test.
+    // - Second possible value is "xvfb" which makes the IDE run in the xvfb mode (using virtual display; only available
+    //   under X Window System). This config can be set either with a raw string value or by setting the IDEPROBE_DISPLAY
```
PawelLipski, line 289: ` This config can be set either with a raw string value or by setting the IDEPROBE_DISPLAY...` applies to the entire `display` config key, not just to the `Second possible value`

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/InstalledIntelliJ.scala
```diff
@@ -36,8 +36,12 @@ sealed abstract class InstalledIntelliJ(root: Path, probePaths: IdeProbePaths, c
   protected val implementationSpecificVmOptions: Seq[String] = Seq.empty
 
   protected lazy val vmoptions: Path = {
+    val shouldRunInHeadlessMode = displayMode match {
+      case Display.Headless => true
+      case _                => false
+    }
```
PawelLipski, line 42:
```suggestion
    val shouldRunInHeadlessMode = displayMode == Display.Headless
```
❓

core/driver/sources/src/main/resources/reference.conf
LukaszKontowski: Yeah, my mistake. Will be fixed 👍

core/driver/sources/src/main/resources/reference.conf
```diff
@@ -301,15 +308,6 @@ probe {
...
     }
 
-    // `driver.headless` defines if IDE should run in a headless mode (without a graphical user interface). If you want
-    // to use it, set to `true`. Specifically, it adds " headless" to the basic launch command ("idea").
-    // Note: if `headless` is set to `true`, it takes precedence over `driver.display`, which means that
-    // IDE will run in the headless mode (and not in a "native" or "xvfb" mode).
-    // Additionally, unless run in headless mode, the driver captures the screenshots of the screen periodically,
-    // so it can then be exposed on the CI for the user to inspect the IDE state in case of failures. If you run in the
-    // headless mode - then no screenshots will be captured.
-    headless = false
```
PawelLipski, line 311: This is a breaking change in general (albeit on a rather rarely used option)... TBD if we want to first deprecate (but maintain support) for `driver.headless` for some time before it's finally removed

LukaszKontowski: I talked with @lukaszwawrzyk about this and his suggestion was that we can just remove this whole feature and follow your suggestion from https://github.com/VirtusLab/ide-probe/issues/306#issuecomment-1295046843
Also, in my opinion, let's remove it and not keep some pointless parts of code/configs. It is not used anyway.

PawelLipski: Ok, +1

PawelLipski (approved):

core/driver/tests/src/test/resources/CustomCommand/ideprobe.conf
```diff
@@ -2,7 +2,7 @@ probe {
   workspace.path = "classpath:/CustomCommand"
 
   driver {
-    headless = false
```
lukaszwawrzyk, line 5: it was false, so it should be "native" or empty, right?

lukaszwawrzyk (approved):

core/driver/tests/src/test/resources/CustomCommand/ideprobe.conf
LukaszKontowski: yes, my mistake - just fixed it

Issue #324, opened 2022-11-20
UI tests stuck at `org.virtuslab.ideprobe.dependencies.IntelliJPatternResolver.replaceGlobsWithExistingDirectories`


PawelLipski:
`jstack` on test process:

![image](https://user-images.githubusercontent.com/3383210/202910557-da8486eb-b9c1-4a24-85b3-c976a423b228.png)

Happens only in CI for some reason (e.g. [this build](https://app.circleci.com/pipelines/github/VirtusLab/git-machete-intellij-plugin/7664/workflows/4aa66772-c0ec-48a0-a5be-33089dfb3303/jobs/8087)):

![image](https://user-images.githubusercontent.com/3383210/202910578-26fee184-8545-44d7-9991-b670c41caf4b.png)

Could not reproduce locally so far

PawelLipski:
The problem is in the tail-recursive implementation of `org.virtuslab.ideprobe.dependencies.IntelliJResolver#replaceGlobsWithExistingDirectories`, which apparently never ends if no match is found.

For the record, the offending glob seems to have been:

```
"file:///"${HOME}"/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/[revision]/*/ideaIC-[revision]/",
```

for `revision = 223.7571.58-EAP-SNAPSHOT`

- [ ] fix the implementation (preferably by replacing the custom logic with an established library for globs)
- [ ] add regression test that would've captured the problem in the first place

PawelLipski:
Simple repro: add the following path in `org.virtuslab.ideprobe.dependencies.IntelliJResolverTest#resolvesIntelliJPatternWithGlobesUsed`:

```scala
    val probeConfig = IntelliJFixture.readIdeProbeConfig(
      Config.fromString(s"""
                           |probe.resolvers.intellij.repositories = [
                           |  "file:///non/*/path",   # this one!
                           |  "$mavenRepo/com/*/intellij/*/$mavenArtifact/${mavenVersion.build}/$mavenArtifact-${mavenVersion.build}.zip"
                           |]
                           |""".stripMargin),
      "probe"
    )
```

PawelLipski:
Out the options I've checked, https://github.com/EsotericSoftware/wildcard seems the most reasonable (other libs, including `java.nio` and apache-commons seem to go through every single file under the specified root folder, which is a 2/10 solution)

PR #325, opened 2022-11-21
Fix glob resolver stalling on non-existent paths

PawelLipski:
Fix for #324

### DIFF
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/DependencyResolver.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/DependencyResolver.scala
@@ -3,3 +3,9 @@ package org.virtuslab.ideprobe.dependencies
 trait DependencyResolver[Key] {
   def resolve(key: Key): Dependency
 }
+
+object DependencyResolver {
+  def apply[Key](f: Key => Dependency): DependencyResolver[Key] = new DependencyResolver[Key] {
+    def resolve(key: Key): Dependency = f(key)
+  }
+}
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
@@ -1,8 +1,8 @@
 package org.virtuslab.ideprobe.dependencies
 
+import java.io.File
+import java.nio.file.Path
 import java.nio.file.Paths
-
-import scala.annotation.tailrec
 
 import org.virtuslab.ideprobe.IntelliJFixture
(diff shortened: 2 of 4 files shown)


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/DependencyResolver.scala
```diff
@@ -3,3 +3,9 @@ package org.virtuslab.ideprobe.dependencies
 trait DependencyResolver[Key] {
   def resolve(key: Key): Dependency
 }
+
```
lukaszwawrzyk, line 6: Why is this needed? We can simply pass function where such trait is expected anyway.

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
```diff
@@ -43,54 +43,40 @@ case class IntelliJPatternResolver(pattern: String) extends IntelliJResolver {
...
-    val replaced =
-      if (replacedBeforeResolvingGlobs.startsWith("file:"))
-        resolveGlobsInPattern(replacedBeforeResolvingGlobs)
-      else
-        replacedBeforeResolvingGlobs
 
-    Dependency(replaced)
+    if (artifactWithReplacementsApplied.startsWith("file:"))
+      findFilesMatchingGlob(artifactWithReplacementsApplied.stripPrefix("file:"))
+        .map("file:" + _)
```
lukaszwawrzyk, line 52: shouldn't we stript `file://`?

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/DependencyResolver.scala
PawelLipski: Isn't needed per se... but makes it much more clear than we construct an instance of `DependencyResolver` 😅 

I was confused when saw

```scala
def resolver(artifact: String): DependencyResolver[IntelliJVersion] = { version =>
....
  Dependency(artifactWithReplacementsApplied)
}
```

Didn't notice the `version =>` and though that `def resolver` returns a `Dependency`, not a `DependencyResolver`

core/driver/sources/src/main/scala/org/virtuslab/ideprobe/dependencies/IntelliJResolver.scala
PawelLipski: Unfortunately, even the URIs returned by `java.lang.Class#getResource` (like, `java.lang.Class#getResource`) are only prefixed with `file:` instead of `file://`... so we probably shouldn't rely on standard being observed here 😅

PR #413, opened 2024-02-23
Store plugins (robot, ide-probe, the tested plugin) in a directory separate from where bundled plugins live

PawelLipski:
https://github.com/VirtusLab/git-machete-intellij-plugin/issues/1820

TL;DR it seems that under IntelliJ 2024.1 EAP 5, the approach with saving the plugins alongside bundled plugins directory no longer works. ideprobe-provided plugins aren't loaded at all, crashing the test on socket connection attempt. The likely cause are the changes applied for [IJPL-633](https://youtrack.jetbrains.com/issue/IJPL-633).

### DIFF
--- a/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJPaths.scala
+++ b/core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJPaths.scala
@@ -14,7 +14,7 @@ final case class IntelliJPaths(
     userPrefs: Path
 ) {
   val bin: Path = root.resolve("bin")
-  val bundledPlugins = root.resolve("plugins")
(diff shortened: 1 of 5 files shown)


core/driver/sources/src/main/scala/org/virtuslab/ideprobe/ide/intellij/IntelliJPaths.scala
```diff
@@ -14,7 +14,7 @@ final case class IntelliJPaths(
     userPrefs: Path
 ) {
   val bin: Path = root.resolve("bin")
-  val bundledPlugins = root.resolve("plugins")
+  val bundledPlugins: Path = root.resolve("plugins")
```
PawelLipski, line 17: It's now only used in

image

